Issue #63, opened 2020-03-31
Read activities


enriquezrene:
Activities screen has been created and it is listing activities hard-coded in a JSON file. We need to revisit this implementation and make the proper request to read these activities using the API.

**Acceptance Criteria**
- Replace the existing implementation to read activities using the following endpoint 
`GET http://timetracker-api.azurewebsites.net/activities`
- Remove the JSON file with activities
- Update the existing unit tests

**Note**
- The API is just a mock but it can work for our purposes

PR #74, opened 2020-04-03
reading activities from API and introducing ngrx #63

enriquezrene:
Reading activities using the api
Removing hard-coded activities in JSON file
Introducing ngrx


macrisguncay:
@enriquezrene I was checking the PR and I think the PR is great..!! I have only one observation,  we are organizing the imports in two sections (the first are angular imports and another section with our imports). Can we fix this, please?

enriquezrene:
Sure thing, on it...

Issue #83, opened 2020-04-08
Remove clock-out/clock-in label


enriquezrene:
Time-clock screen has the following label at the top of the screen
<img width="959" alt="Screen Shot 2020-04-08 at 09 46 52" src="https://user-images.githubusercontent.com/1638636/78797906-f49ca180-797d-11ea-8e5c-956909a9db5c.png">

That information is not useful at all so we need to remove it.

**Acceptance Criteria**
- The above-mentioned section is removed

PR #115, opened 2020-04-13
#83 removing not needed labels from clock-in screen

enriquezrene:
#83 removing not needed labels from clock-in screen


enriquezrene:
closes #83

Issue #85, opened 2020-04-08
List activities from API


enriquezrene:
After selecting a project all the details to register a time-entry are displayed as the following image shows:
<img width="959" alt="Screen Shot 2020-04-08 at 09 54 15" src="https://user-images.githubusercontent.com/1638636/78798705-ed29c800-797e-11ea-9877-c42f71e55a44.png">

Activities are saved in the API and we should list all of them in here using a drop-down component.

**Acceptance Criteria**
- Implement the required logic to get the activities from the API and list them here
- Replace the input text by a drop-down component
- All the code needs to have at least 80% code coverage

PR #116, opened 2020-04-13
#85 reading activities with api

enriquezrene:
#85 reading activities with api

### DIFF
--- a/src/app/modules/time-clock/pages/time-clock.component.ts
+++ b/src/app/modules/time-clock/pages/time-clock.component.ts
@@ -1,3 +1,8 @@
+import { LoadActivities } from '../../activities-management/store/activity-management.actions';
+import { ActivityState } from '../../activities-management/store/activity-management.reducers';
+import { Store, select } from '@ngrx/store';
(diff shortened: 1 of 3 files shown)


src/app/modules/time-clock/pages/time-clock.component.ts
```diff
@@ -1,3 +1,8 @@
+import { LoadActivities } from '../../activities-management/store/activity-management.actions';
+import { ActivityState } from '../../activities-management/store/activity-management.reducers';
+import { Store, select } from '@ngrx/store';
+import { allActivities } from '../../activities-management/store';
+import { Activity } from 'src/app/modules/shared/models';
```
jorgecod, line 5: Please, organize imports by module.
**import { Component, OnInit } from '@angular/core';**
**import { Store, select } from '@ngrx/store';**
**the rest of imports..**

And if there is an index file in the store directory, you can import in one line

enriquezrene: done

Issue #90, opened 2020-04-08
Manage customer project types


enriquezrene:
As part of #87 and #89, we have implemented all the required functionality to manage customers' information. Now, we need to add the required code to manage project types customers have. 

A project type is comprised of:
- Name (required)
- Description (optional)

Considering these fields, we need to implement the required code to create, edit and delete activity types. 

**Acceptance Criteria**
- All the required code to manage activity types has at least 80% code coverage
- User can manage project types

PR #152, opened 2020-04-20
fix: #90 Manage customer project types

DiegoTinitana:


src/app/modules/shared/models/project-type.model.ts
```diff
@@ -0,0 +1,6 @@
+export interface ProjectType {
+  id: string;
+  name: string;
+  description: string;
```
macrisguncay, line 4: description is optional, please add ? symbol.

src/app/modules/shared/models/project-type.model.ts
```diff
@@ -0,0 +1,6 @@
...
+  tenant_id?: string;
```
macrisguncay, line 5: @enriquezrene I'm not sure if tenant_id still should be a property. We can get it from session storage and in this case, we don't use it

src/app/modules/customer-management/components/projects-type/components/create-project-type/create-project-type.component.html
```diff
@@ -1,11 +1,31 @@
...
+      <div class="text-danger" *ngIf="(name.dirty || name.touched) && name.invalid && name.errors.required">
+        Name is required.
+      </div>
+      <textarea
+        class="form-control form-control-sm mt-2"
+        id="descriptionTextArea"
+        rows="3"
+        placeholder="Description"
+        formControlName="description"
+      ></textarea>
```
macrisguncay, line 23: we can set a default value for optional fields, that way we can avoid the error when we do two continue creations only with required fields

Issue #91, opened 2020-04-08
Manage projects


enriquezrene:
As part of #87 and #89, we have implemented all the required functionality to manage customers' information. Now, we need to add the required code to manage projects customers have.

A project is comprised of:
- Name (required)
- Description (optional)
- Type (required)
(Check this image for more info https://user-images.githubusercontent.com/1638636/78589039-3229ef00-7805-11ea-9c2e-64ec07362dc0.png)

Furthermore, we need to add a search box at the top of the table so end-users can find projects easily.

**Acceptance Criteria**
- All the required code needs to have at least 80% code coverage
- The screen should allow creating, editing and removing projects

PR #160, opened 2020-04-22
fix: #91 manage-projects

macrisguncay:
this is the ticket: https://github.com/ioet/time-tracker-ui/issues/91


src/app/modules/customer-management/components/projects/components/services/project.service.ts
```diff
@@ -21,12 +22,16 @@ export class ProjectService {
     return this.http.post<Project[]>(this.url, {
       ...projectData,
       customer_id: 'b6e6a2f1-ce5c-49b3-8649-d154b4f3c305',
-      tenant_id: 'b95385bd-d0f4-4823-b874-af9421c14c8e',
+      tenant_id: this.azureAdB2CService.getTenantId(),
```
enriquezrene, line 25: we no longer need a tenant_id, this value is going to be read from the JWT so feel free to remove it 😃

src/app/modules/customer-management/components/projects/components/create-project/create-project.component.html
```diff
@@ -1,28 +1,44 @@
...
+      <option [value]="" disabled selected>Select project type</option>
+      <option *ngFor="let type of projectsTypes" [value]="type.id">{{ type.name }}</option>
+    </select>
+    <div
+      class="text-danger"
+      *ngIf="
+        (project_type_id.dirty || project_type_id.touched) && project_type_id.invalid && project_type_id.errors.required
+      "
+    >
+      Project type is required.
```
enriquezrene, line 31: Actually, the type is not required. Let's remove this validation.

Issue #147, opened 2020-04-20
Display correct username


enriquezrene:
As a user, I want to see my name when I log-in to the application so that to avoid having confusion during the system usage.

- A user log-in to the application using their credentials
- The right corner should display their username
- When the `Time-clock` option is select it is always displaying the following message that needs to be corrected as well
```
Dario clocked  ... 
```

PR #163, opened 2020-04-23
fix: #147 Display correct username fix test

DiegoTinitana:


src/app/modules/activities-management/components/create-activity/create-activity.component.spec.ts
```diff
@@ -20,6 +21,7 @@ describe('CreateActivityComponent', () => {
   let activityIdtoEditMock;
   let allActivitiesMock;
   let getActivityByIdMock;
+  let getActivityByIdSelectorMock;
```
macrisguncay, line 24: yo can use the variable getActivityByIdMock that is to mock that selector, you don't need to create getActivityByIdSelectorMock

src/app/modules/customer-management/components/projects-type/components/create-project-type/create-project-type.component.spec.ts
```diff
@@ -20,6 +21,7 @@ describe('InputProjectTypeComponent', () => {
   let projectTypeIdToEditMock;
   let allProjectTypesMock;
   let getProjectTypeByIdMock;
+  let getProjectTypeByIdSelectorMock;
```
macrisguncay, line 24: tha same case. you can use getProjectTypeByIdMock and not create a new variable.

PR #235, opened 2020-05-13
222/feature create technologies component

juanultimate:
It extracts the technologies dropdown as a component.


src/app/modules/shared/components/technologies/technologies.component.ts
```diff
@@ -0,0 +1,72 @@
...
+  ngOnInit(): void {
+    const technologies$ = this.store.pipe(select(allTechnologies));
+    technologies$.subscribe((response) => {
+      this.isLoading = response.isLoading;
+      const filteredItems = response.technologyList.items.filter(item => !this.selectedTechnologies.includes(item.name));
+      this.technology = {items: filteredItems};
+    });
+  }
+
+  addTechnology(name: string) {
```
enriquezrene, line 56: We don't want to display already selected technologies in the component @juanultimate. Only not selected technologies should be displayed in the component.

PR #270, opened 2020-05-19
228 table component

juanultimate:
Including Angular datatables on customer list
image


juanultimate:
Tests has been added again

Issue #180, opened 2020-04-24
Show time-entries for selected month


enriquezrene:
As a user, I want to visit the URL https://time-tracker-ui.azurewebsites.net/time-entries and select any month from the top of the screen in order to see all the time-entries I have entered.

Acceptance Criteria:
- A user can click on any month from the top
- After selecting a month, all the time entries belonging to that month are displayed (use this endpoint https://github.com/ioet/time-tracker-backend/issues/106)
- Users can edit, delete any of the displayed time-entries
- The list needs to display
- Time entries are ordered by start_time field and grouped by day
- Edit and delete button need to have same styles than other buttons in the app
_____________________________________________
| start-time | end-time | duration | project |
....

PR #303, opened 2020-05-26
fix: #180 Show time-entries for selected month

DiegoTinitana:


src/app/modules/time-entries/pages/time-entries.component.ts
```diff
@@ -39,8 +31,9 @@ export class TimeEntriesComponent implements OnInit {
   }
 
   editEntry(entryId: string) {
+    console.log(this.dataByMonth);
```
enriquezrene, line 34: Let's remove this line.

Issue #371, opened 2020-06-10
Add confirmation before deleting data


enriquezrene:
We need to add a confirmation message before deleting:
- Activities
- Projects
- Customers
- Project type

PR #387, opened 2020-06-15
fix: #371 Add_confirmation_before_deleting_data

DiegoTinitana:


src/app/modules/shared/components/dialog/dialog.component.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+      activity: 'development',
+      technologies: ['Angular', 'TypeScript'],
+    };
+
+    spyOn(component.removeList, 'emit');
+    component.list = merged;
+    fixture.detectChanges();
+    component.removeListById(merged.id);
+    expect(component.removeList.emit).toHaveBeenCalled();
+    component.cancelDeleteModal.nativeElement.click();
```
juanultimate, line 45: What is this for?

src/app/modules/shared/components/dialog/dialog.component.html
```diff
@@ -0,0 +1,17 @@
+<div class="modal-dialog" role="document">
+  <div class="modal-content" *ngIf="list">
+    <div class="modal-header">
+      <h5 class="modal-title" id="exampleModalLabel">{{ title }}</h5>
+      <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+        <span aria-hidden="true">&times;</span>
+      </button>
+    </div>
+    <div class="modal-body">
+      Are you sure you want to delete <b>{{ list.project_name || list.name }}</b>
```
juanultimate, line 10: Don't we need a question mark at the end of this sentence?

src/app/modules/shared/components/dialog/dialog.component.html
```diff
@@ -0,0 +1,17 @@
+<div class="modal-dialog" role="document">
+  <div class="modal-content" *ngIf="list">
+    <div class="modal-header">
+      <h5 class="modal-title" id="exampleModalLabel">{{ title }}</h5>
```
enriquezrene, line 4: id would me modalTitle or so. Please replace **exampleModalPanel**

enriquezrene: Any chance to put the title using a bold style?

src/app/modules/shared/components/dialog/dialog.component.html
enriquezrene: We do need it for sure. So messages should look like:

Are you sure you want to delete **XYZ**?

src/app/modules/shared/components/dialog/dialog.component.html
```diff
@@ -0,0 +1,17 @@
...
+      <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+        <span aria-hidden="true">&times;</span>
+      </button>
+    </div>
+    <div class="modal-body">
+      Are you sure you want to delete <b>{{ list.project_name || list.name }}</b>
+    </div>
+    <div class="modal-footer">
+      <button (click)="removeListById(list.id)" type="button" class="btn btn-primary">Delete</button>
```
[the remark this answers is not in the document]
enriquezrene: Probably not, since this component is intended to be used just to confirm delete actions. In this sense, I'd rather call it
_src/app/modules/shared/components/dialogs/remove.dialog.component.html_

The functionality works good but we need to move the buttons so the first option is **Cancel** and the second one is **Delete**. This is how confirmation modal panels usually look like:
Screen Shot 2020-06-16 at 09 51 06

Taking the image into account, let's use the style **danger** for the **Delete** button.

src/app/modules/shared/components/dialog/dialog.component.spec.ts
DiegoTinitana: this line is to close de modal

src/app/modules/shared/components/dialog/dialog.component.html
DiegoTinitana: yes this is working, I need to put this {{ list.project_name || list.name }} because the objects don't have the same names.

src/app/modules/shared/components/dialog/dialog.component.html
```diff
@@ -0,0 +1,17 @@
...
+      <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+        <span aria-hidden="true">&times;</span>
+      </button>
+    </div>
+    <div class="modal-body">
+      {{ body }}
+    </div>
+    <div class="modal-footer">
+      <button #cancelDeleteModal type="button" class="btn btn-secondary" data-dismiss="modal">Cancel</button>
+      <button (click)="closeModal()" type="button" class="btn btn-danger">Delete</button>
```
juanultimate, line 14: Replace Delete by Confirm so that this modal can be used by other confirm scenarios

Issue #395, opened 2020-06-16
Make the activity required when clocking-out


Yaneida:
If the user has not entered the Activity data, that field should be highlighted in red because it's required when clocking-out.

PR #411, opened 2020-06-22
fix: #395 Make the activity required when clocking-out

DiegoTinitana:


src/app/modules/time-clock/pages/time-clock.component.ts
```diff
@@ -32,8 +34,16 @@ export class TimeClockComponent implements OnInit {
     });
   }
 
-  clockOut() {
+  stopEntry() {
     this.store.dispatch(new StopTimeEntryRunning(this.activeTimeEntry.id));
     this.areFieldsVisible = false;
   }
+
+  clockOut() {
```
juanultimate, line 42: Any chance we can add unit tests to this new code?

PR #416, opened 2020-06-23
set seconds to zero on create time-entry #399

### LINKED ISSUE #399: set seconds to 00
When a new entry is created from the time-entries menu option we need to set the start_times with seconds 00 and the same for end time.

When a user clock-in we need to set the current time with 00 seconds as well, and when the user clocks out we need to do the same.

You need to visit the functionalities in:
- time-entries.component.ts => doSave(entry) function
- project-list-hover.component.ts => clockIn(selectedProject, customerName, name) function
- project-list-hover.component.ts => switch(selectedProject, customerName, name) function
- time-clock.component.ts =>  clockOut() funcion

Make sure the backend allows saving this info, for example an entry:

start time => 18-06-2020T10:00:00Z
**end time => 18-06-2020T11:00:00Z**

**start time => 18-06-2020T11:00:00Z**
end time => 18-06-2020T15:00:00Z

Notice that both entries have the same value for end_time and start_time

Angeluz-07:
fixes #399

### DIFF
--- a/src/app/modules/time-entries/pages/time-entries.component.ts
+++ b/src/app/modules/time-entries/pages/time-entries.component.ts
@@ -66,17 +66,13 @@ export class TimeEntriesComponent implements OnInit {
     if (this.entryId) {
       const startDateChanged = this.entry.start_date !== event.entry.start_date;
       if (startDateChanged) {
-        const startDate = new Date(event.entry.start_date);
-        startDate.setSeconds(1, 0);
-        event.entry.start_date = startDate.toISOString();
+        event.entry.start_date = this.adjustDateSecs(event.entry.start_date, 1);
       }
 
       if (endDateIsDefined) {
         const endDateChanged = this.entry.end_date !== event.entry.end_date;
         if (endDateChanged) {
-          const endDate = new Date(event.entry.end_date);
-          endDate.setSeconds(0, 0);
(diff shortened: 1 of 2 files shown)


src/app/modules/time-entries/pages/time-entries.component.ts
```diff
@@ -62,17 +62,40 @@ export class TimeEntriesComponent implements OnInit {
...
-    event.entry.start_date = new Date(event.entry.start_date).toISOString();
-    if (event.entry.end_date !== null && event.entry.end_date !== undefined) {
-      event.entry.end_date = new Date(event.entry.end_date).toISOString();
-    }
     if (this.entryId) {
+      const startDateChanged = this.entry.start_date !== event.entry.start_date;
+      const endDateChanged = this.entry.end_date !== event.entry.end_date;
+
+      if (startDateChanged) {
+        const startDate = new Date(event.entry.start_date);
```
juanultimate, line 70: Lines 71, 72, 73 are the same than 89,90,91. Do not you think it would be great if we [extract them to a method](https://refactoring.guru/extract-method).

Angeluz-07: Your're right. I thought the same. I just didn't know the code style to include functions. I will ask for help to Rene for this and also to fix the tests. Thanks for pointing it out.

Issue #459, opened 2020-07-02
Don't allow starting a time entry in the future


astridschneider:
A time entry cannot be started in the future. Please validate. 
(Right now you can start a time entry in the future if you select "I'm working on this".

PR #466, opened 2020-07-07
fix: validate future date #459

Angeluz-07:


src/app/modules/shared/components/details-fields/details-fields.component.ts
```diff
@@ -168,6 +169,13 @@ export class DetailsFieldsComponent implements OnChanges, OnInit {
     if (this.goingToWorkOnThis) {
       delete entry.end_date;
     }
+
+    const invalidEntryDate = new Date(entryDate) > new Date();
```
juanultimate, line 173: I would name this variables as `isEntryDateInTheFuture`

PR #468, opened 2020-07-09
:construction: Send timezone_offset when time_entries are being created/updated

### LINKED ISSUE #464: Send timezone_offset when time_entries are being created/updated
When a time-entry is being created we need to send the timezone_offset as part of the payload.

Consider this requirement for:
- Time-clock page (clock-in, switch)
- Time-entries page (creation, edition)

This ticket needs to be implemented after https://github.com/ioet/time-tracker-backend/issues/190

Angeluz-07:
fixes #464

### DIFF
--- a/src/app/modules/time-clock/store/entry.effects.ts
+++ b/src/app/modules/time-clock/store/entry.effects.ts
@@ -21,7 +21,11 @@ export class EntryEffects {
         map((response) => {
           const stopDateForEntry = new Date(response.end_date);
           stopDateForEntry.setSeconds(stopDateForEntry.getSeconds() + 1 );
-          return new actions.CreateEntry({ project_id: action.idProjectSwitching, start_date: stopDateForEntry.toISOString() });
+          return new actions.CreateEntry({
+            project_id: action.idProjectSwitching,
+            start_date: stopDateForEntry.toISOString(),
+            timezone_offset: new Date().getTimezoneOffset(),
+          });
         }),
         catchError((error) => {
           this.toastrService.warning(error.error.message);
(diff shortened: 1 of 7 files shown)


src/app/modules/time-clock/store/entry.effects.ts
```diff
@@ -21,7 +21,11 @@ export class EntryEffects {
         map((response) => {
           const stopDateForEntry = new Date(response.end_date);
           stopDateForEntry.setSeconds(stopDateForEntry.getSeconds() + 1 );
-          return new actions.CreateEntry({ project_id: action.idProjectSwitching, start_date: stopDateForEntry.toISOString() });
+          return new actions.CreateEntry({
+            project_id: action.idProjectSwitching,
+            start_date: stopDateForEntry.toISOString(),
```
juanultimate, line 26: I know this is not your change, but it seems something is wrong here:
**start_date**: **stopDateForEntry**.toISOString(),

Angeluz-07: This has to do with the logic of clocking out, when switching a project. The stopDate of the entry being clocked out is the startDate of the new time-entry.

juanultimate: Yep, I get what you mean. In that case I would change line 23 to this:
`const newTimeEntryStartDate = stopDateForEntry.setSeconds(stopDateForEntry.getSeconds() + 1 );`
and line  26 to:
`start_date: newTimeEntryStartDate.toISOString(),`

In that way, it is easier to read.

Issue #453, opened 2020-07-01
Include guids for people and customer in reports table


enriquezrene:
The data coming from the application is being used to be aggregated with data from other applications so we need to provide users a consistent way to identify them so we need to display the guids (PKs) for customers and projects in the reports table.

Acceptance Criteria
- Display the customer id and project id columns in the reports page
- Make sure these columns are included in the generated report

PR #472, opened 2020-07-20
fix: #453 include customer and project id in reports

enriquezrene:
Please don't merge this PR before this PR is merged and deployed https://github.com/ioet/time-tracker-backend/pull/197

### DIFF
--- a/src/app/modules/shared/models/entry.model.ts
+++ b/src/app/modules/shared/models/entry.model.ts
@@ -9,6 +9,8 @@ export interface Entry {
   project_id?: string;
   owner_email?: string;
   description?: string;
+  customer_id?: string;
+  customer_name?: string;
 }
 
 export interface NewEntry {
--- a/src/app/modules/reports/components/time-entries-table/time-entries-table.component.html
+++ b/src/app/modules/reports/components/time-entries-table/time-entries-table.component.html
@@ -7,6 +7,9 @@
         <th class="col sm-col">Date</th>
         <th class="col sm-col">Duration</th>
         <th class="col md-col">Project</th>
+        <th class="col lg-col">Project ID</th>
+        <th class="col md-col">Customer</th>
+        <th class="col lg-col">Customer ID</th>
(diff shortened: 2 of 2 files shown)


src/app/modules/shared/models/entry.model.ts
```diff
@@ -9,6 +9,8 @@ export interface Entry {
   project_id?: string;
   owner_email?: string;
   description?: string;
+  customer_id?: string;
```
juanultimate, line 12: When are these fields going to be null or undefined?

enriquezrene: When you are creating an entry.

enriquezrene: The time_entry table has the project id information, the customer info is not saved as part of it.

juanultimate: True fact. However, from my perspective, the **time_entry** table is encapsulated into the API. The frontend does not ( and should not ) know about the database data structures. The frontend does ( and should ) know only about the data structures the endpoints are using ( which might be different from the database structures). In this case, the [Entry](https://github.com/ioet/time-tracker-ui/blob/master/src/app/modules/shared/models/entry.model.ts#L1) model/data structure/DTO represents the payload received from the `/time-entries` endpoint. Therefore, the customer and project information is not going to be empty since there is no way to create an entry without specifying the project ( and thus the customer).

juanultimate: In addition, I think we can improve this data structure using compound objects. Something like this:
```
interface Entry {
  id: string;
  running: boolean;
  start_date: Date;
  end_date?: Date;
  activity_id?: string;
  technologies: string[];
  uri?: string;
  project :{
    id: String
    name: String
   customer:{
      id: String
      name: String
  },
  ...... 
}
```

enriquezrene: Yup, good idea

PR #538, opened 2020-10-14
fix: make responsive forms of activities section #537

### LINKED ISSUE #537: Fix forms of activities section
Subtask of #529

Angeluz-07:
close #537

### DIFF
--- a/src/app/modules/activities-management/components/activity-list/activity-list.component.scss
+++ b/src/app/modules/activities-management/components/activity-list/activity-list.component.scss
@@ -1,5 +1,5 @@
 @import '../../../../../styles/colors.scss';
 
 .activity-list {
-  overflow: auto; display:inline-block; width: 60%;
+  overflow: auto;
 }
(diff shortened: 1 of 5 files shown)


src/app/modules/activities-management/pages/activities-management.component.scss
```diff
@@ -0,0 +1,9 @@
+.activity-container{
+    width: 100%;
```
juanultimate, line 2: bootstrap has a w-100 class which does this

Angeluz-07: Yes bootstrap has those classes, but I created this class to use the media-query. On desktop screen we need 80% of width, but in phone we need 100%. Kind of style conditionally.

Angeluz-07: I am not sure if can do it, only with bootstrap classes, without using empty divs.

Angeluz-07: You gave me the idea. I could do it with bootstrap classes only. Thanks

PR #564, opened 2020-11-05
feat: #562 Show read only GUID field

### LINKED ISSUE #562: Show read only GUID field in customers and project sections.
Add guid of project in the following section. Same for customers.

image

PaulRC-ioet:
Show read-only GUID field in customers and project sections.

Close #562

### DIFF
--- a/src/app/modules/customer-management/components/customer-info/components/create-customer/create-customer.html
+++ b/src/app/modules/customer-management/components/customer-info/components/create-customer/create-customer.html
@@ -1,6 +1,6 @@
 <div class="container">
   <form style="width: 600px;" [formGroup]="customerForm" (ngSubmit)="onSubmit(customerForm.value)">
-    <div class="form-group">
+      <div class="form-group">
       <input
         class="form-control form-control-sm"
         id="customerName"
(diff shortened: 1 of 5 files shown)


Angeluz-07 (approved):

src/app/modules/customer-management/components/customer-info/components/create-customer/create-customer.html
```diff
@@ -1,6 +1,17 @@
 <div class="container">
   <form style="width: 600px;" [formGroup]="customerForm" (ngSubmit)="onSubmit(customerForm.value)">
     <div class="form-group">
+      <div class="form-group">
```
enriquezrene, line 4: This is not going to solve the problem, what we need to do instead is to show the uuid in the table. Please add this column in the customer list, project list, and project type list as well.

enriquezrene (changes requested):

src/app/modules/customer-management/components/customer-info/components/create-customer/create-customer.html
PaulRC-ioet: Got it, Working on that!

PaulRC-ioet:
All the changes are done @enriquezrene , delete de input with the ids, and add those ids into the tables.

enriquezrene (approved):

PR #582, opened 2020-12-04
TT-15 keep the user logged in

gqroot:
Manage user session in the browser to avoid logging in to the app again

### DIFF
--- a/src/app/guards/login-guard/login.guard.spec.ts
+++ b/src/app/guards/login-guard/login.guard.spec.ts
@@ -1,7 +1,6 @@
 import { TestBed, inject } from '@angular/core/testing';
 import { RouterTestingModule } from '@angular/router/testing';
 import { Router } from '@angular/router';
-import { CookieService } from 'ngx-cookie-service';
 
 import { AzureAdB2CService } from '../../modules/login/services/azure.ad.b2c.service';
 import { LoginGuard } from './login.guard';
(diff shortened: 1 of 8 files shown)


src/app/modules/login/services/azure.ad.b2c.service.ts
```diff
@@ -27,6 +34,8 @@ export class AzureAdB2CService {
   }
 
   logout() {
+    this.cookieService.delete('msal.idtoken');
+    this.cookieService.delete('msal.idtoken');
```
enriquezrene, line 38: should we do it twice?

src/app/guards/login-guard/login.guard.spec.ts
```diff
@@ -24,16 +26,19 @@ describe('LoginGuard', () => {
...
     azureAdB2CService = TestBed.inject(AzureAdB2CService);
+    cookieService = TestBed.inject(CookieService);
   });
 
   it('should be created', () => {
     expect(loginGuard).toBeTruthy();
   });
 
-  it('can activate the route when user is logged-in', () => {
+  it('can activate the route when user is logged-in && the token cookie exists', () => {
```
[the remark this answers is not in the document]
gqroot: It is only possible to do 3 of the 4 tests. Translate the logic of the condition to the `AzureAdB2CService` class

src/app/modules/login/services/azure.ad.b2c.service.ts
gqroot: No, I already corrected it, the one to be deleted is the 2nd cookie

josepato87 (approved):

PR #584, opened 2020-12-10
TT-69 Overlap Hours

gqroot:
This PR is the 1st part of the task, the 2nd part is in the backend

### DIFF
--- a/src/app/modules/time-entries/pages/time-entries.component.ts
+++ b/src/app/modules/time-entries/pages/time-entries.component.ts
@@ -100,15 +100,17 @@ export class TimeEntriesComponent implements OnInit, OnDestroy {
   }
 
   saveEntry(event: SaveEntryEvent): void {
-    if (this.activeTimeEntry !== null && this.activeTimeEntry !== undefined) {
+    if (this.activeTimeEntry) {
       const startDateAsLocalDate = new Date(event.entry.start_date);
       const endDateAsLocalDate = new Date(event.entry.end_date);
       const activeEntryAsLocalDate = new Date(this.activeTimeEntry.start_date);
-      const isEditingActiveEntry = this.entryId === this.activeTimeEntry.id;
-      if (!isEditingActiveEntry && (startDateAsLocalDate > activeEntryAsLocalDate || endDateAsLocalDate > activeEntryAsLocalDate)) {
-        this.toastrService.error('You are on the clock and this entry overlaps it, try with earlier times.');
-      } else {
+      const isEditingEntryEqualToActiveEntry = this.entryId === this.activeTimeEntry.id;
+      const isStartDateLowerThanActiveEntry = startDateAsLocalDate < activeEntryAsLocalDate;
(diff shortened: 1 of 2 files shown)


src/app/modules/time-entries/pages/time-entries.component.ts
```diff
@@ -101,11 +101,11 @@ export class TimeEntriesComponent implements OnInit, OnDestroy {
 
   saveEntry(event: SaveEntryEvent): void {
     if (this.activeTimeEntry !== null && this.activeTimeEntry !== undefined) {
-      const entryDateAsIso = new Date(event.entry.start_date).toISOString();
-      const entryDateAsLocalDate = new Date(entryDateAsIso);
+      const startDateAsLocalDate = new Date(event.entry.start_date);
+      const endDateAsLocalDate = new Date(event.entry.end_date);
       const activeEntryAsLocalDate = new Date(this.activeTimeEntry.start_date);
       const isEditingActiveEntry = this.entryId === this.activeTimeEntry.id;
```
josepato87, line 107: Not sure if `isEditingActiveEntry` is the right description of that comparison???

src/app/modules/time-entries/pages/time-entries.component.ts
```diff
@@ -101,11 +101,11 @@ export class TimeEntriesComponent implements OnInit, OnDestroy {
...
-      if (!isEditingActiveEntry && entryDateAsLocalDate > activeEntryAsLocalDate) {
+      if (!isEditingActiveEntry && (startDateAsLocalDate > activeEntryAsLocalDate || endDateAsLocalDate > activeEntryAsLocalDate)) {
```
josepato87, line 108: Maybe changing the order of the statement would be more understandable.

`if(true) {saves}
else {show error}`

josepato87, line 108: Don't hesitate to use more variables instead of using operators inside the conditions. For example:
`isStartDateGreaterThanActiveEntry = startDateAsLocalDate > activeEntryAsLocalDate
isEndDateGreaterThanActiveEntry = endDateAsLocalDate > activeEntryAsLocalDate`
Then the condition would be more readable. 

Besides, due to the previous comment, about changing the order of the statement. The result would be something like this:

```
isStartDateLowerThanActiveEntry = startDateAsLocalDate < activeEntryAsLocalDate
isEndDateLowerThanActiveEntry = endDateAsLocalDate < activeEntryAsLocalDate
if (isEditingActiveEntry && (isStartDateLowerThanActiveEntry || isEndDateLowerThanActiveEntry)) {
  this.doSave(event);
} 
else {
  this.toastrService.error('You are on the clock and this entry overlaps it, try with earlier times.');
}
```

josepato87 (commented):
Just some refactoring suggestions

src/app/modules/time-entries/pages/time-entries.component.ts
gqroot: maybe `isEditingEntryEqualToActiveEntry` ?

josepato87: Better :)

PR #591, opened 2020-12-17
TT-85 fix: fix overlapping seconds on time entries

PaulRC-ioet:
fix a bug that not allow to update changes on "time entries".

### DIFF
--- a/src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
+++ b/src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
@@ -319,7 +319,7 @@ describe('DetailsFieldsComponent', () => {
     expect(component.saveEntry.emit).toHaveBeenCalledWith(data);
   });
 
-  fit('onSubmit an entry without change hours, should not modify the start_date and end_date ', () => {
+  it('onSubmit an entry without change hours, should not modify the start_date and end_date ', () => {
     component.entryToEdit = {...entryToEdit, description: 'test', };
     fixture.componentInstance.ngOnChanges();
 
--- a/src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
+++ b/src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
@@ -316,22 +316,22 @@ describe('DetailsFieldsComponent', () => {
   });
 
   it('should not modify the start_date when start_hour has not been modified', () => {
-    const dateTest = moment().format('YYYY-MM-DD');
-    const startHourTest = moment().subtract(3, 'hours').format('HH:mm:ss');
-    const expectedStartDate = new Date(`${dateTest}T${startHourTest.trim()}`);
+    const currentDate = moment().format('YYYY-MM-DD');
+    const startHour = moment().subtract(3, 'hours').format('HH:mm:ss');
+    const expectedStartDate = new Date(`${currentDate}T${startHour.trim()}`);
 
     component.entryToEdit = {...entryToEdit, start_date: expectedStartDate };
     fixture.componentInstance.ngOnChanges();
 
     component.entryForm.patchValue({ description: 'test' });
 
-    expect(component.startDateToSubmit()).toEqual(expectedStartDate);
+    expect(component.dateToSubmit('start_date', 'start_hour')).toEqual(expectedStartDate);
   });
 
   it('should modify the start_date when start_hour has been modified', () => {
-    const dateTest = moment().format('YYYY-MM-DD');
-    const startHourTest = moment().format('HH:mm:ss');
-    const startDate = new Date(`${dateTest}T${startHourTest.trim()}`);
+    const currentDate = moment().format('YYYY-MM-DD');
+    const startHour = moment().format('HH:mm:ss');
+    const startDate = new Date(`${currentDate}T${startHour.trim()}`);
 
     component.entryToEdit = {...entryToEdit, start_date: startDate };
     fixture.componentInstance.ngOnChanges();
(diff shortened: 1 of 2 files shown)


src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
```diff
@@ -76,6 +76,11 @@ describe('DetailsFieldsComponent', () => {
     description: '',
     technology: '',
   };
+  const dateTest = moment().format('YYYY-MM-DD');
+  const endHourTest = moment().format('HH:mm:ss');
+  const endDateTest = new Date(`${dateTest}T${endHourTest.trim()}`);
+  const startHourTest = moment().subtract(3, 'hours').format('HH:mm:ss');
+  const startDateTest = new Date(`${dateTest}T${startHourTest.trim()}`);
```
josepato87, line 83: I would move these two lines before the `end` variables

josepato87, line 83: Do you think we could have any issue about the timing where the tests run? For example, if we put this test running at 1:00 am, due you are subtracting 3 hours, it means that the current hour would be 23:00 of what day?

src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
```diff
@@ -314,6 +319,19 @@ describe('DetailsFieldsComponent', () => {
...
+    component.entryToEdit = {...entryToEdit, description: 'test', };
+    fixture.componentInstance.ngOnChanges();
+
+    const startHourValue = moment().subtract(3, 'hours').format('HH:mm');
+    const endHourValue = moment().format('HH:mm');
+
+    component.onSubmit();
+
+    expect(component.starDateValue).toEqual(startHourValue);
+    expect(component.endDateValue).toEqual(endHourValue);
```
josepato87, line 332: How are you asserting that the other fields, like description are being updated?

src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
```diff
@@ -314,6 +319,19 @@ describe('DetailsFieldsComponent', () => {
     expect(component.saveEntry.emit).toHaveBeenCalledWith(data);
   });
 
+  it('onSubmit an entry without change hours, should not modify the start_date and end_date ', () => {
+    component.entryToEdit = {...entryToEdit, description: 'test', };
+    fixture.componentInstance.ngOnChanges();
+
+    const startHourValue = moment().subtract(3, 'hours').format('HH:mm');
+    const endHourValue = moment().format('HH:mm');
```
josepato87, line 327: not sure if this should be the best approach for setting the start and end hours values, as I understand, you are setting those values in the `before each` section. Could you read the current entry "hours" and setting the test variables with those values instead?

src/app/modules/shared/components/details-fields/details-fields.component.ts
```diff
@@ -204,13 +208,20 @@ export class DetailsFieldsComponent implements OnChanges, OnInit {
     }
     const startDate = this.entryForm.value.start_date;
     const endDate = this.entryForm.value.end_date;
+    this.starDateValue = this.entryForm.value.start_hour === this.starDateValue ?
+      this.entryToEdit.start_date :
+      new Date(`${startDate}T${this.entryForm.value.start_hour.trim()}`).toISOString();
+    this.endDateValue = this.entryForm.value.end_hour === this.endDateValue ?
+      this.entryToEdit.end_date :
+      new Date(`${endDate}T${this.entryForm.value.end_hour.trim()}`).toISOString();
```
josepato87, line 216: I see an opportunity to extract lines 213 and 216 into a method that receives the `entryForm.value.X_VALUE` and returns everything that is being done in that large sentence.

If you find that in other files, could be an utilitary. :)

src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
PaulRC-ioet: I have change all the tests, please check them.

PaulRC-ioet: I have deleted that block of lines, it's different now.

src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
PaulRC-ioet: there's a new test to prove that.

src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
PaulRC-ioet: I have also deleted those hours, now is in the test those values.

src/app/modules/shared/components/details-fields/details-fields.component.ts
PaulRC-ioet: It's Done thanks.

src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
```diff
@@ -310,13 +312,71 @@ describe('DetailsFieldsComponent', () => {
       },
       shouldRestartEntry: false
     };
-
     expect(component.saveEntry.emit).toHaveBeenCalledWith(data);
   });
 
+  it('should not modify the start_date when start_hour has not been modified', () => {
+    const dateTest = moment().format('YYYY-MM-DD');
```
scastillo-jp, line 319: change: 
 `dateTest` to `currentDate`
 `startHourTest` to `startHour`

I think is necessary use other name of variable per example  `dateTest` to `dateCurrent`

PaulRC-ioet: Ok it's changed

src/app/modules/shared/components/details-fields/details-fields.component.ts
```diff
@@ -197,29 +197,56 @@ export class DetailsFieldsComponent implements OnChanges, OnInit {
...
     }
-    const isStartDateInTheFuture = moment(startDate).isAfter(moment());
-    const isEndDateInTheFuture = moment(endDate).isAfter(moment());
-    if (isStartDateInTheFuture || isEndDateInTheFuture) {
+
+    const isStartDateInTheFuture = moment(startDateToSubmit).isAfter(moment());
+    const isEndDateInTheFuture = moment(endDateToSubmit).isAfter(moment());
+    const timeEntryIsInTheFuture = isStartDateInTheFuture || isEndDateInTheFuture;
+
+    if (timeEntryIsInTheFuture) {
```
Angeluz-07, line 249: I remember we made a change to extract `timeEntryIsInTheFuture` into a function. I guess that caused issues in existing tests and you ended up doing it like this. Am I right, or  is it an error?

PaulRC-ioet: I just tried the function of not allowing a "time entry" in the future and it allowed me to add. It ignores the way we put it so I left it as it was. but with the tests, there was no problem.

Angeluz-07: It's ok. Then, keep in mind that, when we may want to retake that little refactor.

josepato87 (approved):
LGTM

PR #592, opened 2020-12-21
TT-64 fix: set technologies as list when create time-entry

Angeluz-07:
With these changes, we ensure that at the moment the time-entry is created, the technologies field is set as an empty list. See TT-64 for details.


josepato87 (approved):
LGTM, please don't forget that we have some tests pending :)

PR #596, opened 2020-12-30
TT-39 fix: default month when changing year

gqroot:


PaulRC-ioet (approved):
It is a good way to refactor the code, in my opinion, everything is fine.

PR #601, opened 2021-01-04
TT-108 fix: edit the first running entry of the month after one hour

gqroot:


src/app/modules/shared/components/month-picker/month-picker.component.spec.ts
```diff
@@ -62,7 +62,7 @@ describe('MonthPickerComponent', () => {
   it('selectMonth should call selectDate', () => {
     spyOn(component, 'selectDate');
     component.selectMonth(10);
-    expect(component.selectDate).toHaveBeenCalledWith(10, 2020);
+    expect(component.selectDate).toHaveBeenCalledWith(10, 2021);
```
PaulRC-ioet, line 65: What would happen next year? would we have to change? Wouldn't it be better to add a moment with a current year?

gqroot: you are right, i have already corrected it

josepato87 (approved):

PR #604, opened 2021-01-05
TT-107 feat: Limit the year selector in UI to the current year

scastillo-jp:
Screen Shot 2021-01-06 at 10 55 01 AM

Screen Shot 2021-01-06 at 10 54 37 AM

https://user-images.githubusercontent.com/12575632/103682550-0afc7b80-4f57-11eb-91cf-75ee1b9d5990.mov

### DIFF
--- a/src/app/modules/shared/components/month-picker/month-picker.component.html
+++ b/src/app/modules/shared/components/month-picker/month-picker.component.html
@@ -1,14 +1,21 @@
 <div class="card-header">
-  <div class="d-flex align-items-center">
-    <button class="btn mr-auto" (click)="changeYear('subtract')">
-      <i class="fa fa-chevron-left"></i>
-    </button>
-    <p>
-      {{selectedYearText}}
-    </p>
-    <button class="btn ml-auto" (click)="changeYear('add')">
-      <i class="fa fa-chevron-right"></i>
-    </button>
+  <div class="row align-items-center">
+    <div class="col-sm-4">
+      <button class="btn mr-auto" (click)="changeYear('subtract')">
+        <i class="fa fa-chevron-left"></i>
+      </button>
+    </div>
+    <div class="col-sm-4">
+      <p class="text-center year-text">
+        {{selectedYearText}}
+      </p>
+    </div>
+
+    <div class="col-sm-4 text-right">
+      <button *ngIf="selectedYear < currentYear" class="btn ml-auto" (click)="changeYear('add')">
+        <i class="fa fa-chevron-right"></i>
+      </button>
+    </div>
   </div>
 </div>
 <div class="row content-months">
--- a/src/app/modules/shared/components/month-picker/month-picker.component.html
+++ b/src/app/modules/shared/components/month-picker/month-picker.component.html
@@ -1,25 +1,26 @@
 <div class="card-header">
-  <div class="row align-items-center">
-    <div class="col-sm-4">
+  <div class="row">
+    <div class="col-4">
       <button class="btn mr-auto" (click)="changeYear('subtract')">
         <i class="fa fa-chevron-left"></i>
       </button>
     </div>
-    <div class="col-sm-4">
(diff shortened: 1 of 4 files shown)


src/app/modules/shared/components/month-picker/month-picker.component.spec.ts
```diff
@@ -61,8 +61,8 @@ describe('MonthPickerComponent', () => {
 
   it('selectMonth should call selectDate', () => {
     spyOn(component, 'selectDate');
-    component.selectMonth(10);
-    expect(component.selectDate).toHaveBeenCalledWith(10, 2020);
+    component.selectMonth(1);
+    expect(component.selectDate).toHaveBeenCalledWith(1, 2021);
```
josepato87, line 65: I have some comments about this. We have a deep dependency on the current year. Maybe the solution done by Guido would be better: https://github.com/ioet/time-tracker-ui/pull/601#discussion_r552101062

src/app/modules/shared/components/month-picker/month-picker.component.html
```diff
@@ -1,14 +1,21 @@
...
+      </button>
+    </div>
+    <div class="col-sm-4">
+      <p class="text-center year-text">
+        {{selectedYearText}}
+      </p>
+    </div>
+
+    <div class="col-sm-4 text-right">
+      <button *ngIf="selectedYear < currentYear" class="btn ml-auto" (click)="changeYear('add')">
```
josepato87, line 15: Maybe we would need a test for this. I don't know how hard would be to test in this file, or maybe we could put the logic in the TS file?

src/app/modules/shared/components/month-picker/month-picker.component.spec.ts
scastillo-jp: @josepato87  the implementation made to Guido are failing, other option is get the the current month an year to test

src/app/modules/shared/components/month-picker/month-picker.component.html
scastillo-jp: I go to check it.

src/app/modules/shared/components/month-picker/month-picker.component.spec.ts
scastillo-jp: another option is to get a dynamic month and year to not use static values for example

  fit('selectMonth should call selectDate', () => {
    spyOn(component, 'selectDate');
    const monthSelect = new Date().getMonth();
    const yearSelect = new Date().getFullYear()
    component.selectMonth(monthSelect);
    expect(component.selectDate).toHaveBeenCalledWith(monthSelect, yearSelect);
  });

israteneda: I think it's a better approach since if the current year stay static, it'll be necessary to change it every year.

Angeluz-07 (approved):

src/app/modules/shared/components/month-picker/month-picker.component.ts
```diff
@@ -40,6 +41,8 @@ export class MonthPickerComponent implements OnInit {
 
     const monthIndex = changeAction === 'add' ? 0 : 11;
     this.selectMonth(monthIndex);
+
+    this.showArrowNext = this.selectedYear < this.currentYear ?? !this.showArrowNext;
```
josepato87, line 45: I won't put `!this.showArrowNext` I'll just put `false` is better for reading.

scastillo-jp: ok, the default value of `this.showArrowNext` is false, in this case I have to change to` true`

israteneda (approved):

israteneda (approved):

PaulRC-ioet (approved):

josepato87 (approved):

PR #614, opened 2021-01-13
TT-111 fix: Add a roles field in Users

PaulRC-ioet:
Add a user Roles field in the model User

### DIFF
--- a/src/app/modules/users/components/users-list/users-list.component.spec.ts
+++ b/src/app/modules/users/components/users-list/users-list.component.spec.ts
@@ -69,7 +69,7 @@ describe('UsersListComponent', () => {
     expect(component.users).toEqual(state.data);
   });
 
-  it('on success load users, the data of role should be an array and role null', () => {
+  it('on success load users, the data of roles should be an array and role null', () => {
     const actionSubject = TestBed.inject(ActionsSubject) as ActionsSubject;
     const action = {
       type: UserActionTypes.LOAD_USERS_SUCCESS,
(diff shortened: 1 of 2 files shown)


israteneda (approved):

src/app/modules/users/models/users.ts
```diff
@@ -2,6 +2,7 @@ export interface User {
   name: string;
   email: string;
   role?: string;
+  roles?: string[];
```
josepato87, line 5: Where are the tests for this? I mean, when we saw the implementation you showed me a simulation of FF so you used a mocked users json, you could test in the same way by mocking the response of the endpoint and what the model would receive. Is it that possible?

josepato87 (commented):
Testing, testing please

src/app/modules/users/models/users.ts
PaulRC-ioet: Sure, I have added a test to check it

src/app/modules/users/components/users-list/users-list.component.spec.ts
```diff
@@ -59,6 +69,21 @@ describe('UsersListComponent', () => {
     expect(component.users).toEqual(state.data);
   });
 
+  it('on success load users, the data of role should be an array and role null', () => {
```
josepato87, line 72: ...the data of roles... (not role)

josepato87, line 72: Maybe you could add a new test with the inverse case of role and roles... so you can modify the `data` on line 17 to have info in `role`, and `roles` as empty

josepato87 (commented):
Left some quick solving comments

src/app/modules/users/components/users-list/users-list.component.spec.ts
```diff
@@ -14,7 +14,17 @@ describe('UsersListComponent', () => {
   const actionSub: ActionsSubject = new ActionsSubject();
 
   const state: UserState = {
-    data: [{ name: 'name', email: 'email', role: 'role', id: 'id', tenant_id: 'tenant id', deleted: 'delete' }],
+    data: [
+      {
+        name: 'name',
+        email: 'email',
+        role: null,
+        roles: ['admin', 'test'],
```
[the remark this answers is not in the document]
josepato87: That's temporary. We will remove `role` when the whole feature around `roles` is completed. Right now is under a FF in the backend, but the frontend would accept both for now without any issues in case we switch to use one or the other field.

israteneda (approved):

src/app/modules/users/components/users-list/users-list.component.spec.ts
juanultimate: Gotcha!!

Angeluz-07: The inverse case should have `role` as string and `roles` as null.

PaulRC-ioet: Done.

PaulRC-ioet: Ok, I have tested that case also.

josepato87 (approved):

src/app/modules/users/components/users-list/users-list.component.spec.ts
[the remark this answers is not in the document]
juanultimate: Got it. It would be great if you add a TODO and link to the issue which is going to remove the role attribute.

PR #627, opened 2021-01-19
Tt 117 two entries in progress bug

scastillo-jp:


src/app/modules/time-entries/pages/time-entries.component.ts
```diff
@@ -145,9 +146,6 @@ export class TimeEntriesComponent implements OnInit, OnDestroy {
     if (this.entryId) {
       event.entry.id = this.entryId;
       this.store.dispatch(new entryActions.UpdateEntry(event.entry));
-      if (event.shouldRestartEntry) {
-        this.store.dispatch(new entryActions.RestartEntry(event.entry));
-      }
```
josepato87, line 150: I'm not sure why we remove the restart. Not an issue in the PR, but I would like to understand why we removed this :)

src/app/modules/time-entries/pages/time-entries.component.ts
```diff
@@ -104,17 +104,18 @@ export class TimeEntriesComponent implements OnInit, OnDestroy {
   }
 
   saveEntry(event: SaveEntryEvent): void {
-    if (this.activeTimeEntry) {
+
+    if (this.activeTimeEntry && this.entryId === this.activeTimeEntry.id) {
       const startDateAsLocalDate = new Date(event.entry.start_date);
       const endDateAsLocalDate = new Date(event.entry.end_date);
       const activeEntryAsLocalDate = new Date(this.activeTimeEntry.start_date);
-      const isEditingEntryEqualToActiveEntry = this.entryId === this.activeTimeEntry.id;
```
josepato87, line 111: I wouldn't remove this variable... I would put at the top of the method and use it instead of the second validation in the first if clause

src/app/modules/time-entries/pages/time-entries.component.spec.ts
```diff
@@ -247,7 +247,7 @@ describe('TimeEntriesComponent', () => {
         uri: 'abc',
       }, shouldRestartEntry: false
     };
-    component.entryId = 'new-entry';
+    component.entryId = 'entry_1';
```
josepato87, line 250: I'm not sure why we changed this value... entry_1 is another entry used in other tests.

src/app/modules/time-entries/pages/time-entries.component.spec.ts
```diff
@@ -363,12 +364,13 @@ describe('TimeEntriesComponent', () => {
...
       }, shouldRestartEntry: true
     };
+
     component.entryId = '123';
     spyOn(store, 'dispatch');
 
     component.doSave(entryToSave);
 
-    expect(store.dispatch).toHaveBeenCalledWith(new entryActions.RestartEntry(entryToSave.entry));
+    expect(store.dispatch).toHaveBeenCalledWith(new entryActions.UpdateEntry(entryToSave.entry));
```
josepato87, line 373: The same as the other comment, just wan to understand why we change from "Restart..." to "Update..."

josepato87 (commented):
I believe that creating tests scenarios for active entries and other entries should work better.

src/app/modules/time-entries/pages/time-entries.component.ts
```diff
@@ -145,9 +146,6 @@ export class TimeEntriesComponent implements OnInit, OnDestroy {
     if (this.entryId) {
       event.entry.id = this.entryId;
       this.store.dispatch(new entryActions.UpdateEntry(event.entry));
-      if (event.shouldRestartEntry) {
-        this.store.dispatch(new entryActions.RestartEntry(event.entry));
```
PaulRC-ioet, line 149: If we eliminate the Restart action, would we no longer use that endpoint?

src/app/modules/time-entries/pages/time-entries.component.spec.ts
scastillo-jp: @josepato87 the if of `entry_1` is from  `component.activeTimeEntry = entry;` line 239

src/app/modules/time-entries/pages/time-entries.component.spec.ts
scastillo-jp: With the new changes it is not necessary to restart the entry when I edit the entries.
Only with the update does not get the error of two inputs in progress

src/app/modules/time-entries/pages/time-entries.component.ts
josepato87: What happened to this comment? I see that you just get rid of the variable and you are not longer validating it. Is it because we don't need to verify if the `this.activeTimeEntry` is the same as `this.entryId`?

src/app/modules/time-entries/pages/time-entries.component.spec.ts
josepato87: Yep, but if you consider the purpose of the test is to throw error when something is wrong... we don't need to change the value... `new-entry` is ok for the context of the test, I believe.

src/app/modules/time-entries/pages/time-entries.component.spec.ts
```diff
@@ -267,7 +267,7 @@ describe('TimeEntriesComponent', () => {
         uri: 'abc',
       }, shouldRestartEntry: false
     };
-    component.entryId = 'new-entry';
+    component.entryId = 'entry_1';
```
josepato87, line 271: The same as before regarding the id name

src/app/modules/time-entries/pages/time-entries.component.spec.ts
josepato87: Mmm if you say that you don't need to `restart`, why are you setting `shouldRestartEntry` as `true` in this test?
Please consider a review on the whole test.

src/app/modules/shared/components/details-fields/details-fields.component.ts
```diff
@@ -126,6 +126,7 @@ export class DetailsFieldsComponent implements OnChanges, OnInit {
 
   ngOnChanges(): void {
     this.goingToWorkOnThis = this.entryToEdit ? this.entryToEdit.running : false;
+    this.shouldRestartEntry = false;
```
josepato87, line 129: In general it looks good to me... Just wondering (I don't remember) if there is any test for the case when `this.shouldRestartEntry` is `true`. If not, that would be something to implement, please.

josepato87 (commented):
LGTM, just left a comment regarding missing test

src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
```diff
@@ -300,14 +302,6 @@ describe('DetailsFieldsComponent', () => {
     expect(component.goingToWorkOnThis).toBeFalse();
   });
 
-  it('when editing entry that already finished, then the entry should not be marked as running', () => {
```
Angeluz-07, line 303: This tests seems to be useful. Why do we deleted?

scastillo-jp: @Angeluz-07 this test are repeated.

josepato87 (approved):

israteneda (approved):

PR #632, opened 2021-01-27
fix: TT-138 The table users does not adapt to mobile screens

LEON12699:
The table users is responsive for mobile screens

### DIFF
--- a/src/app/modules/users/components/users-list/users-list.component.html
+++ b/src/app/modules/users/components/users-list/users-list.component.html
@@ -1,5 +1,5 @@
 <div class="table-responsive">
-    <table
+  <table
     *ngIf="users"
     class="table table-sm table-bordered table-striped mb-0"
     datatable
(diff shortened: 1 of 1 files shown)


src/app/modules/users/components/users-list/users-list.component.html
```diff
@@ -1,38 +1,40 @@
...
+              size="small"
+              (change)="switchRole(user.id, user.roles, 'test', 'time-tracker-tester')"
+              [checked]="user.roles.includes('time-tracker-tester')"
+            ></ui-switch>
+            test
+          </div>
+        </td>
+      </tr>
+    </tbody>
+    </table>
```
scastillo-jp, line 39: @LEON12699 delete two spaces, before of `</table >`

israteneda (approved):

israteneda (approved):

scastillo-jp (approved):

src/app/modules/users/components/users-list/users-list.component.html
```diff
@@ -1,38 +1,40 @@
...
-            (change)="switchRole(user.id, user.roles, 'test', 'time-tracker-tester')"
-            [checked]="user.roles.includes('time-tracker-tester')"
-          ></ui-switch>
-          test
-        </div>
-      </td>
-    </tr>
-  </tbody>
-</table>
+<div class="table-responsive">
```
josepato87, line 1: Is this really enough for making this table responsive?

LEON12699: Yes, in addition "flex-wrap" and "text-break"  are the only necessary things to make this table responsive

PR #635, opened 2021-02-02
fix: TT-144 Responsive all tables

jr-98:
All tables are now responsive

### DIFF
--- a/src/app/modules/reports/components/time-entries-table/time-entries-table.component.scss
+++ b/src/app/modules/reports/components/time-entries-table/time-entries-table.component.scss
@@ -26,3 +26,10 @@
 .hidden-col{
   display: none;
 }
+.table {
+  overflow-x: auto;
+}
+.row{
+  width: 100%;
+  display: grid;
+}
--- a/src/app/modules/reports/components/time-entries-table/time-entries-table.component.scss
+++ b/src/app/modules/reports/components/time-entries-table/time-entries-table.component.scss
@@ -26,10 +26,9 @@
 .hidden-col{
   display: none;
 }
-.table {
-  overflow-x: auto;
-}
-.row{
+
+.scroll-table {
+  overflow-x: scroll;
   width: 100%;
   display: grid;
 }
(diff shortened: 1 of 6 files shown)


juanultimate (approved):

src/app/modules/time-entries/pages/time-entries.component.html
```diff
@@ -25,7 +25,7 @@
         <th class="col">Customer</th>
         <th class="col">Project</th>
         <th class="col">Activity</th>
-        <th class="col"></th>
+        <th class="col">Options</th>
```
juanultimate, line 28: This change may broke unit tests

jr-98: ok, I will reverted this change

jr-98: OK. I will reverted this change

src/app/modules/reports/components/time-entries-table/time-entries-table.component.scss
```diff
@@ -26,3 +26,8 @@
 .hidden-col{
   display: none;
 }
+.table.dataTable th,
```
scastillo-jp, line 29: @jr-98 is this style working? dataTable is not a class

jr-98: I took as an example a table that had a similar problem and the solution was that and it worked? What should I change?

jr-98: @scastillo-jp  The class does work

josepato87 (approved):

scastillo-jp (approved):

scastillo-jp:
@jr-98 
I found this problem when the data is loading
Screen Shot 2021-02-05 at 11 12 25 AM

After loading the data, the table does not have the correct margin: 
Screen Shot 2021-02-05 at 11 12 34 AM

scastillo-jp:
Uploading Screen Recording 2021-02-18 at 3.56.46 PM.mov…

scastillo-jp (approved):

PR #636, opened 2021-02-03
fix: TT-140 time entry info is lost if project is selected at last

LEON12699:
It is necessary to choose project to enter the data

### DIFF
--- a/src/app/modules/shared/components/details-fields/details-fields.component.html
+++ b/src/app/modules/shared/components/details-fields/details-fields.component.html
@@ -1,13 +1,20 @@
 <form [formGroup]="entryForm" (ngSubmit)="onSubmit()">
-  <label *ngIf='canMarkEntryAsWIP'><input id='isEntryRunning' type="checkbox" (change)="onGoingToWorkOnThisChange($event)" [checked]="goingToWorkOnThis"> I am working on
-    this</label>
+  <label *ngIf="canMarkEntryAsWIP">
+    <input
+      id="isEntryRunning"
+      type="checkbox"
+      (change)="onGoingToWorkOnThisChange($event)"
+      [checked]="goingToWorkOnThis"
+    />
+    I am working on this
+  </label>
 
   <div class="form-group row">
     <label class="col-12 col-sm-2 col-form-label">Project:</label>
     <div class="col-12 col-sm-10 autocomplete">
       <ng-autocomplete
-        (selected)='onSelectedProject($event)'
-        (inputCleared)='onClearedComponent($event)'
+        (selected)="onSelectedProject($event)"
+        (inputCleared)="onClearedComponent($event)"
         formControlName="project_name"
         [data]="listProjects"
         [searchKeyword]="keyword"
@@ -15,7 +22,9 @@
         notFoundText="No projects found"
         placeHolder="Enter the project name"
         [itemTemplate]="itemTemplate"
-        [notFoundTemplate]="notFoundTemplate">
+        [notFoundTemplate]="notFoundTemplate"
+        [class.valid]="project_name.invalid"
+      >
       </ng-autocomplete>
 
       <ng-template #itemTemplate let-item>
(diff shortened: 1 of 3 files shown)


scastillo-jp (approved):

scastillo-jp:
@here this change looks good,  lost if project is selected at last,  good job @Israel León Gálvez  @William Bravo TT-140 time entry info is 
Just one change adds a border radius at the input for the entry / exit time. so that everything looks similar.
Screen Shot 2021-02-05 at 11 45 54 AM

src/app/modules/shared/components/details-fields/details-fields.component.html
```diff
@@ -1,21 +1,30 @@
...
         formControlName="project_name"
         [data]="listProjects"
         [searchKeyword]="keyword"
         historyIdentifier="projectsSelected"
         notFoundText="No projects found"
         placeHolder="Enter the project name"
         [itemTemplate]="itemTemplate"
-        [notFoundTemplate]="notFoundTemplate">
+        [notFoundTemplate]="notFoundTemplate"
+        [class.valid]="project_name.invalid"
```
Angeluz-07, line 26: I know I said to add a feedback. But looking at the UI, It doesn't look ok for me at first. Let's remove this line `[class.valid]="project_name.invalid"` and think another way of feedback.

image

src/app/modules/shared/components/details-fields/details-fields.component.html
```diff
@@ -34,108 +43,112 @@
...
-      >
-        <option value="" selected="selected"></option>
-        <option *ngFor="let activity of activities" value="{{ activity.id }}">{{ activity.name }}</option>
-      </select>
+  <fieldset [disabled]="!(project_id.value && project_name.value)">
+    <div class="form-group row">
+      <label class="col-12 col-sm-2 col-form-label">Activity:</label>
+      <div class="col-12 col-sm-10">
+        <select
+          [class.is-invalid]="activity_id.invalid"
```
Angeluz-07, line 51: I know this is not part of your ticket. But let's improve the activity field feedback by changing this line to `[class.is-invalid]="activity_id.invalid && activity_id.touched"`

src/app/modules/shared/components/details-fields/details-fields.component.scss
```diff
@@ -93,3 +93,9 @@ input[type="date"]::-webkit-clear-button {
   border: none;
   margin: 0em;
 }
+
+.valid ::ng-deep .autocomplete-container ::ng-deep .input-container ::ng-deep input  {
```
Angeluz-07, line 97: What would be the purpose of this new style?

Angeluz-07 (changes requested):

src/app/modules/shared/components/details-fields/details-fields.component.html
Angeluz-07: We can also discuss it with the team, to get a consensus.

src/app/modules/shared/components/details-fields/details-fields.component.scss
LEON12699: It is the only way to change the style of the "ngx-autocomplete" component without affecting the default style of the component. It's use only for border-color of the input

src/app/modules/shared/components/details-fields/details-fields.component.ts
```diff
@@ -246,6 +246,12 @@ export class DetailsFieldsComponent implements OnChanges, OnInit {
     this.saveEntry.emit({ entry, shouldRestartEntry: this.shouldRestartEntry });
   }
 
+  onclickFormAction(isProjectSelected: boolean){
+    if (isProjectSelected){
+      this.toastrService.warning('Please, select first a project');
```
Angeluz-07, line 251: I think we could change the message to "Please, first select a project"

LEON12699: Ok i´m going to  change it

Angeluz-07 (approved):
Well done.

scastillo-jp (approved):

PR #642, opened 2021-02-26
fix: TT-167 Date picker disable future dates

wobravo:
disable future dates on date in and date out

### DIFF
--- a/src/app/modules/shared/components/details-fields/details-fields.component.html
+++ b/src/app/modules/shared/components/details-fields/details-fields.component.html
@@ -91,6 +91,8 @@
             [class.is-invalid]="start_date.invalid && start_date.touched"
             required
             (ngModelChange)="onStartDateChange($event)"
+            [max]="getActualDate()"
+            onkeydown="return false"
           />
         </div>
 
@@ -118,6 +120,8 @@
             aria-describedby="inputGroup-sizing-sm"
             [class.is-invalid]="end_date.invalid && end_date.touched"
             required
+            [max]="getActualDate()"
+            onkeydown="return false"
           />
         </div>
 
--- a/src/app/modules/shared/components/details-fields/details-fields.component.html
+++ b/src/app/modules/shared/components/details-fields/details-fields.component.html
@@ -91,7 +91,7 @@
             [class.is-invalid]="start_date.invalid && start_date.touched"
             required
             (ngModelChange)="onStartDateChange($event)"
-            [max]="getActualDate()"
+            [max]="getCurrentDate()"
             onkeydown="return false"
           />
         </div>
@@ -120,7 +120,7 @@
             aria-describedby="inputGroup-sizing-sm"
             [class.is-invalid]="end_date.invalid && end_date.touched"
             required
-            [max]="getActualDate()"
+            [max]="getCurrentDate()"
             onkeydown="return false"
           />
         </div>
--- a/src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
+++ b/src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
@@ -468,6 +468,12 @@ describe('DetailsFieldsComponent', () => {
     expect(endDateInput.value).not.toEqual(startDateInput.value);
     expect(startDateInput.value).toEqual(expectedStartDate);
   });
+
+  it('on get actual date should return the actual date in isoString ', () => {
+    const actualDateExpected = new Date().toISOString().split('T')[0];
+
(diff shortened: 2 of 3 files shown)


src/app/modules/shared/components/details-fields/details-fields.component.html
```diff
@@ -118,6 +120,8 @@
             aria-describedby="inputGroup-sizing-sm"
             [class.is-invalid]="end_date.invalid && end_date.touched"
             required
+            [max]="getActualDate()"
+            onkeydown="return false"
```
israteneda, line 124: Only a question, what does this line do?

israteneda (approved):

src/app/modules/shared/components/details-fields/details-fields.component.html
wobravo: this line does not allow you to write the dates you can only select with the calendar

wobravo: because if that line doesn't go you could write future dates

src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
```diff
@@ -468,6 +468,12 @@ describe('DetailsFieldsComponent', () => {
     expect(endDateInput.value).not.toEqual(startDateInput.value);
     expect(startDateInput.value).toEqual(expectedStartDate);
   });
+
+  it('on get actual date should return the actual date in isoString ', () => {
```
scastillo-jp, line 472: I think we need proof: when we select dates in the future.

src/app/modules/shared/components/details-fields/details-fields.component.ts
```diff
@@ -171,6 +171,10 @@ export class DetailsFieldsComponent implements OnChanges, OnInit {
     this.selectedTechnologies = $event;
   }
 
+  getActualDate(): string {
```
Angeluz-07, line 174: I think you should change this to `getCurrentDate`. And update all references from `actual date` to `current date`

src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
```diff
@@ -468,6 +468,12 @@ describe('DetailsFieldsComponent', () => {
     expect(endDateInput.value).not.toEqual(startDateInput.value);
     expect(startDateInput.value).toEqual(expectedStartDate);
   });
+
+  it('on get actual date should return the current date', () => {
```
Angeluz-07, line 472: I think this test is not that useful. A better test case would be a user trying to change the `start_date` or `end_date` and then,  expecting the app to validate that It can't be a date after the current date.

wobravo: The test is like this because future dates are blocked and the function only gets me the current day.
and the user can only select maximum up to the current date

Angeluz-07: I think it is ok if you want to test function `getActualDate`. But I would recommend to add the other test case.

src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
```diff
@@ -468,6 +468,31 @@ describe('DetailsFieldsComponent', () => {
...
```
Angeluz-07, line 472: We should update the references and the name of the test. Something like:
`On get current date should return expected date`

src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
```diff
@@ -468,6 +468,31 @@ describe('DetailsFieldsComponent', () => {
...
+    const actualDateExpected = new Date().toISOString().split('T')[0];
+
+    expect(component.getCurrentDate()).toEqual(actualDateExpected);
+  });
+
+  it('on the input with id #start_date we could get the id and max value', () => {
+   fixture.detectChanges();
+   const expectDate = new Date().toISOString().split('T')[0];
+   const slectInput: HTMLInputElement = fixture.debugElement.
```
Angeluz-07, line 481: I think if you want to retrieve the selecInput. You only need to queryselector by the id.

Angeluz-07: Also, the name `slectInput` should be `startDateInput` or `dateInput`

scastillo-jp (approved):

scastillo-jp (approved):

Angeluz-07 (approved):

PR #644, opened 2021-03-09
fix: TT-168 Add the parameters startDate and endDate in the path

VanessaIniguez:
Add the parameters startDate and endDate in the path used by the findEntriesByProjectId method of the EntryService service.

### DIFF
--- a/src/app/modules/time-clock/services/entry.service.spec.ts
+++ b/src/app/modules/time-clock/services/entry.service.spec.ts
@@ -143,10 +143,13 @@ describe('EntryService', () => {
 
   it('entries are found by project id with a limit 2 by default', () => {
     const projectId = 'project-id';
+    const startDate = (moment().subtract(1, 'months')).format();
+    const endDate = moment().format();
 
     service.findEntriesByProjectId(projectId).subscribe();
 
-    const restartEntryRequest = httpMock.expectOne( `${service.baseUrl}?limit=2&project_id=${projectId}`);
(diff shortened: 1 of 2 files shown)


kellycastrof:
Yes, but locally the tests were successful.

Angeluz-07 (approved):

src/app/modules/time-clock/services/entry.service.ts
```diff
@@ -77,4 +80,14 @@ export class EntryService {
...
   }
+
+  getDateLastMonth() {
+    return (moment().subtract(1, 'months')).format();
+  }
+
+  getCurrentDate() {
+    return moment().format();
+  }
+
```
scastillo-jp, line 91: remove one of the black line

src/app/modules/time-clock/services/entry.service.spec.ts
```diff
@@ -143,10 +143,13 @@ describe('EntryService', () => {
 
   it('entries are found by project id with a limit 2 by default', () => {
     const projectId = 'project-id';
+    const startDate = (moment().subtract(1, 'months')).format();
```
scastillo-jp, line 146: I have a question.
What is the default format in this `format()` part?
I think you can add a specific format:  example .format('MM/DD/YYYY');  .format('MMM')

VanessaIniguez: The format used to send start_date and end_date in the path is 2021-03-10T15:42:13-05:00,  at the moment I think it should be sent like this

scastillo-jp (approved):

thegreatyamori (approved):

kellycastrof (approved):

LEON12699 (approved):

PR #646, opened 2021-03-11
feat: TT-35 make new entries creation modal panel draggable

eduardisrael:
add @angular/cdk and a unit test

### DIFF
--- a/src/app/modules/time-entries/pages/time-entries.component.spec.ts
+++ b/src/app/modules/time-entries/pages/time-entries.component.spec.ts
@@ -443,10 +443,16 @@ describe('TimeEntriesComponent', () => {
     expect(component.entry).toEqual(newEntry);
   });
 
-  it('When I stop dragging the modal, it should call dragEnded', () => {
-    const dragEndEvent = new DragEvent('CdkDragEnd');
-    spyOn(component, 'dragEnded');
-    component.dragEnded(dragEndEvent);
-    expect(component.dragEnded).toHaveBeenCalledWith(dragEndEvent);
+  it('When I stop dragging the modal, it should reset Draggable Position', () => {
+    const dragEndEventStub = {
+      source: {
+        _dragRef: {
+          reset: () => { }
+        }
+      }
+    };
+    spyOn(dragEndEventStub.source._dragRef, 'reset');
+    component.resetDraggablePosition(dragEndEventStub);
+    expect(dragEndEventStub.source._dragRef.reset).toHaveBeenCalled();
   });
 });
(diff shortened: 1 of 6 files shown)


thegreatyamori:
@eduardisrael a question.
The panel moves freely or is the movement restricted as here?
https://material.angular.io/cdk/drag-drop/overview#restricting-movement-within-an-element

src/app/modules/time-entries/pages/time-entries.component.ts
```diff
@@ -165,4 +165,7 @@ export class TimeEntriesComponent implements OnInit, OnDestroy {
     this.message = `Are you sure you want to delete ${item.activity_name}?`;
     this.showModal = true;
   }
-}
\ No newline at end of file
+  dragEnded(event: any): void {
```
Angeluz-07, line 168: Maybe, we could call this method as `resetDraggablePosition` or something similar.

src/app/modules/time-entries/pages/time-entries.component.spec.ts
```diff
@@ -442,4 +442,11 @@ describe('TimeEntriesComponent', () => {
     component.newEntry();
     expect(component.entry).toEqual(newEntry);
   });
+
+  it('When I stop dragging the modal, it should call dragEnded', () => {
+    const dragEndEvent = new DragEvent('CdkDragEnd');
+    spyOn(component, 'dragEnded');
+    component.dragEnded(dragEndEvent);
+    expect(component.dragEnded).toHaveBeenCalledWith(dragEndEvent);
+  });
```
Angeluz-07, line 451: What about checking the position of the draggable(before and after call the dragEnded function). It was possible or not?

kellycastrof:
> @eduardisrael a question.
> The panel moves freely or is the movement restricted as here?
> https://material.angular.io/cdk/drag-drop/overview#restricting-movement-within-an-element

It is not restricted but if you drop it returns to the initial site.

src/app/modules/time-entries/pages/time-entries.component.ts
eduardisrael: Is it a good idea or it can be resetDragPosition

src/app/modules/time-entries/pages/time-entries.component.spec.ts
eduardisrael: Right now I am trying to verify the draggable positions and create a new unit test if it is the case.

Angeluz-07 (approved):

src/app/modules/time-entries/pages/time-entries.component.spec.ts
eduardisrael: Update the unit test to verify that the reset method is being called

thegreatyamori (approved):

PR #649, opened 2021-03-18
fix: TT-178 Make URI clickable when possible

wobravo:
make uri clickable when possible

### DIFF
--- a/src/app/modules/reports/components/time-entries-table/time-entries-table.component.html
+++ b/src/app/modules/reports/components/time-entries-table/time-entries-table.component.html
@@ -4,7 +4,7 @@
     datatable
     [dtTrigger]="dtTrigger"
     [dtOptions]="dtOptions"
-    *ngIf="(reportDataSource$ | async) as dataSource">
+    *ngIf="reportDataSource$ | async as dataSource">
     <thead class="thead-blue">
       <tr class="d-flex">
         <th class="hidden-col">ID</th>
@@ -41,7 +41,13 @@
         <td class="col md-col">{{ entry.customer_name }}</td>
         <td class="hidden-col">{{ entry.customer_id }}</td>
         <td class="col md-col">{{ entry.activity_name }}</td>
-        <td class="col lg-col">{{ entry.uri }}</td>
+        <td class="col lg-col">
+          <ng-container *ngIf="entry.uri !== null">
+            <a [class.is-url]="isURL(entry.uri)" (click)="openURLInNewTab(entry.uri)">
+              {{ entry.uri }}
+            </a>
+          </ng-container>
+        </td>
         <td class="col lg-col">{{ entry.description }}</td>
         <td class="col lg-col">{{ entry.technologies }}</td>
       </tr>
--- a/src/app/modules/reports/components/time-entries-table/time-entries-table.component.html
+++ b/src/app/modules/reports/components/time-entries-table/time-entries-table.component.html
@@ -4,7 +4,7 @@
     datatable
     [dtTrigger]="dtTrigger"
     [dtOptions]="dtOptions"
-    *ngIf="reportDataSource$ | async as dataSource">
+    *ngIf="(reportDataSource$ | async) as dataSource">
     <thead class="thead-blue">
       <tr class="d-flex">
         <th class="hidden-col">ID</th>
--- a/src/app/modules/reports/components/time-entries-table/time-entries-table.component.scss
+++ b/src/app/modules/reports/components/time-entries-table/time-entries-table.component.scss
@@ -1,3 +1,4 @@
+@import '../../../../../styles/colors.scss';
 .col{
   white-space: nowrap;
   overflow: hidden;
@@ -32,3 +33,25 @@
   width: 100%;
   display: grid;
 }
+
+$url-base-color: $primary;
+$url-hover-color: darken($url-base-color, 20);
+
+.is-url {
+  cursor: pointer;
+  color: $url-base-color;
+  text-decoration: underline;
+
+  &:visited {
+    color: $dark;
+  }
+
+  &:hover {
+    color: $url-hover-color;
+  }
+
+  &:active {
+    color: $url-base-color;
+  }
+}
+
--- a/src/app/modules/reports/components/time-entries-table/time-entries-table.component.spec.ts
+++ b/src/app/modules/reports/components/time-entries-table/time-entries-table.component.spec.ts
@@ -21,7 +21,7 @@ describe('Reports Page', () => {
       description: 'any comment',
       uri: 'custom uri',
       project_id: '123',
-      project_name: 'Time-Tracker'
+      project_name: 'Time-Tracker',
     };
 
     const state: EntryState = {
@@ -33,38 +33,41 @@ describe('Reports Page', () => {
       timeEntriesSummary: null,
       timeEntriesDataSource: {
         data: [timeEntry],
-        isLoading: false
+        isLoading: false,
       },
       reportDataSource: {
         data: [timeEntry],
-        isLoading: false
-      }
+        isLoading: false,
+      },
     };
 
-    beforeEach(waitForAsync(() => {
-      TestBed.configureTestingModule({
-        imports: [],
-        declarations: [TimeEntriesTableComponent, SubstractDatePipe],
-        providers: [provideMockStore({ initialState: state })],
-      }).compileComponents();
-      store = TestBed.inject(MockStore);
+    beforeEach(
+      waitForAsync(() => {
+        TestBed.configureTestingModule({
+          imports: [],
+          declarations: [TimeEntriesTableComponent, SubstractDatePipe],
+          providers: [provideMockStore({ initialState: state })],
+        }).compileComponents();
+        store = TestBed.inject(MockStore);
+      })
+    );
 
-    }));
-
-    beforeEach(waitForAsync(() => {
-      fixture = TestBed.createComponent(TimeEntriesTableComponent);
-      component = fixture.componentInstance;
-      store.setState(state);
-      getReportDataSourceSelectorMock = store.overrideSelector(getReportDataSource, state.reportDataSource);
-      fixture.detectChanges();
-    }));
+    beforeEach(
+      waitForAsync(() => {
+        fixture = TestBed.createComponent(TimeEntriesTableComponent);
+        component = fixture.componentInstance;
+        store.setState(state);
+        getReportDataSourceSelectorMock = store.overrideSelector(getReportDataSource, state.reportDataSource);
+        fixture.detectChanges();
+      })
+    );
 
     it('component should be created', async () => {
       expect(component).toBeTruthy();
     });
 
     it('on success load time entries, the report should be populated', () => {
-      component.reportDataSource$.subscribe(ds => {
+      component.reportDataSource$.subscribe((ds) => {
         expect(ds.data).toEqual(state.reportDataSource.data);
       });
     });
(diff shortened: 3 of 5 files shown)


src/app/modules/reports/components/time-entries-table/time-entries-table.component.scss
```diff
@@ -31,4 +31,26 @@
   overflow-x: scroll;
   width: 100%;
   display: grid;
-}
\ No newline at end of file
+}
+
+$base-color: #00BAEE;
+$hover-color: darken($base-color, 20);
```
Angeluz-07, line 37: If these variables are only valid and useful for the .is-url class, you may want to rename them to something like `url-base-color` & `url-hover-color`.

src/app/modules/reports/components/time-entries-table/time-entries-table.component.ts
```diff
@@ -91,4 +85,12 @@ export class TimeEntriesTableComponent implements OnInit, OnDestroy, AfterViewIn
       this.dtTrigger.next();
     }
   }
+
+  isLinkExternal(uri: string): WindowProxy | string {
```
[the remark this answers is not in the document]
Angeluz-07: Or something like `openLinkInNewTab`.

src/app/modules/reports/components/time-entries-table/time-entries-table.component.ts
```diff
@@ -91,4 +85,12 @@ export class TimeEntriesTableComponent implements OnInit, OnDestroy, AfterViewIn
...
+    return this.isValidUri(uri) ? window.open(uri, '_blank') : '';
+  }
+
+  isValidUri(uri: string) {
```
Angeluz-07, line 93: I think all URIs are valid(at least by now, because we are not validating when the user types them). What we are checking is if the URI is a URL. A better name could be `isURL`

ref: https://danielmiessler.com/study/difference-between-uri-url/

src/app/modules/reports/components/time-entries-table/time-entries-table.component.spec.ts
```diff
@@ -76,6 +79,24 @@ describe('Reports Page', () => {
...
     });
 
+    it('when the uri starts with http or https it should return true and open the url in a new tab', () => {
+      const uriExpected = 'http://customuri.com';
+      spyOn(component, 'isValidUri').and.returnValue(true);
+      spyOn(window, 'open');
+
+      expect(component.isLinkExternal(uriExpected)).not.toEqual('');
+      expect(window.open).toHaveBeenCalledWith(uriExpected, '_blank');
+      expect(component.isValidUri).toHaveBeenCalled();
```
Angeluz-07, line 89: This expect should be called first. So, if it is not called, the test fails just there :)

src/app/modules/reports/components/time-entries-table/time-entries-table.component.spec.ts
```diff
@@ -76,6 +79,24 @@ describe('Reports Page', () => {
...
+    });
+
+    it('when the uri starts without http or https it should return false and not navigate or open a new tab', () => {
+      const uriExpected = timeEntry.uri;
+      spyOn(component, 'isValidUri').and.returnValue(false);
+
+      expect(component.isLinkExternal(uriExpected)).toEqual('');
+      expect(component.isValidUri).toHaveBeenCalled();
```
Angeluz-07, line 97: Same here. This expect should be called first.

src/app/modules/reports/components/time-entries-table/time-entries-table.component.spec.ts
```diff
@@ -76,6 +79,24 @@ describe('Reports Page', () => {
       expect(component.dtTrigger.next).toHaveBeenCalled();
     });
```
Angeluz-07, line 81: It would be useful, if we add another test, to test only the `isValidUri` function. This test could be parametrized and could use the following values:
```
url = 'http://example.com' ; expected_value = true
url = 'https://example.com' ; expected_value = true
url = 'no-url-example' ; expected_value = false
```

src/app/modules/reports/components/time-entries-table/time-entries-table.component.spec.ts
wobravo: Which of the 3 goes first the one of the isValidUri

src/app/modules/reports/components/time-entries-table/time-entries-table.component.spec.ts
```diff
@@ -76,6 +79,24 @@ describe('Reports Page', () => {
...
 
+    it('when the uri starts with http or https it should return true and open the url in a new tab', () => {
+      const uriExpected = 'http://customuri.com';
```
Angeluz-07, line 83: In this test. I think it is not needed to cal the variable `uriExpected` as it is not being compared with other value, it may be just `url`, for readability.

src/app/modules/reports/components/time-entries-table/time-entries-table.component.html
```diff
@@ -1,10 +1,6 @@
 <div class="row scroll-table mt-5 ml-0">
-  <table
-    class="table table-striped mb-0"
-    datatable
```
Angeluz-07, line 4: Could you please restore this multiline format, it is more readable than the two lines.

Angeluz-07 (changes requested):
Good job. Only some fixes.

src/app/modules/reports/components/time-entries-table/time-entries-table.component.spec.ts
Angeluz-07: This `expect(component.isValidUri).toHaveBeenCalled()`

src/app/modules/reports/components/time-entries-table/time-entries-table.component.ts
```diff
@@ -91,4 +85,12 @@ export class TimeEntriesTableComponent implements OnInit, OnDestroy, AfterViewIn
...
+    return uri.startsWith('http' || 'https');
```
LEON12699, line 94: I think this could be `uri.startsWith('http://' || 'https://');` or use a regular expression to evaluate that

src/app/modules/reports/components/time-entries-table/time-entries-table.component.ts
```diff
@@ -24,38 +23,33 @@ export class TimeEntriesTableComponent implements OnInit, OnDestroy, AfterViewIn
...
         columns: ':not(.hidden-col)',
-
       },
       'print',
       {
         extend: 'excel',
         exportOptions: {
           format: {
             body: (data, row, column, node) => {
-              return column === 3 ?
```
thegreatyamori, line 35: can you restore to this indentitation. It is more readable

src/app/modules/reports/components/time-entries-table/time-entries-table.component.ts
```diff
@@ -24,38 +23,33 @@ export class TimeEntriesTableComponent implements OnInit, OnDestroy, AfterViewIn
...
         text: 'Excel',
-        filename: `time-entries-${formatDate(new Date(), 'MM_dd_yyyy-HH_mm', 'en')}`
+        filename: `time-entries-${formatDate(new Date(), 'MM_dd_yyyy-HH_mm', 'en')}`,
       },
       {
         extend: 'csv',
         exportOptions: {
           format: {
             body: (data, row, column, node) => {
-              return column === 3 ?
```
thegreatyamori, line 49: same here.
 Can you restore to this indentation?

src/app/modules/reports/components/time-entries-table/time-entries-table.component.html
```diff
@@ -4,7 +4,7 @@
     datatable
     [dtTrigger]="dtTrigger"
     [dtOptions]="dtOptions"
-    *ngIf="(reportDataSource$ | async) as dataSource">
+    *ngIf="reportDataSource$ | async as dataSource">
```
scastillo-jp, line 7: @wobravo I think it  is necessary  to restore the original changes

src/app/modules/reports/components/time-entries-table/time-entries-table.component.ts
scastillo-jp, line 35: @wobravo restore it  with the original changes

src/app/modules/reports/components/time-entries-table/time-entries-table.component.ts
scastillo-jp, line 49: restore it with the original changes

src/app/modules/reports/components/time-entries-table/time-entries-table.component.spec.ts
```diff
@@ -76,6 +79,34 @@ describe('Reports Page', () => {
...
+      const url = 'http://customuri.com';
+      spyOn(component, 'isURL').and.returnValue(true);
+      spyOn(window, 'open');
+
+      expect(component.openURLInNewTab(url)).not.toEqual('');
+      expect(window.open).toHaveBeenCalledWith(url, '_blank');
+    });
+
+    it('when the uri starts without http or https it should return false and not navigate or open a new tab', () => {
+      const uriExpected = timeEntry.uri;
```
thegreatyamori, line 92: In this part the name would be `url` again

src/app/modules/reports/components/time-entries-table/time-entries-table.component.spec.ts
```diff
@@ -76,6 +79,34 @@ describe('Reports Page', () => {
...
+      expect(component.openURLInNewTab(uriExpected)).toEqual('');
+    });
+
+    const params = [
+      {url: 'http://example.com', expected_value: true, with: 'with'},
+      {url: 'https://example.com', expected_value: true, with: 'with'},
+      {url: 'no-url-example', expected_value: false, with: 'without'}
+    ];
+    params.map((param) => {
+      it(`when the url starts ${param.with} http or https it should return ${param.expected_value}`, () => {
```
Angeluz-07, line 104: A better test name could be `Given the url ${params.url}, the method isURL should return ${params.expected_value}`. In that way, you get rid of the `with` parameter.

src/app/modules/reports/components/time-entries-table/time-entries-table.component.ts
```diff
@@ -1,19 +1,18 @@
 import { formatDate } from '@angular/common';
-import { AfterViewInit, Component, OnDestroy, OnInit, ViewChild } from '@angular/core';
+import { AfterViewInit, Component, OnDestroy, OnInit, ViewChild, NgModule } from '@angular/core';
```
Angeluz-07, line 2: I see you added this NgModule, but I dont see that you use it.

src/app/modules/reports/components/time-entries-table/time-entries-table.component.ts
```diff
@@ -91,4 +85,13 @@ export class TimeEntriesTableComponent implements OnInit, OnDestroy, AfterViewIn
       this.dtTrigger.next();
     }
   }
+
+  openURLInNewTab(uri: string): WindowProxy | string {
```
Angeluz-07, line 89: Nice!

src/app/modules/reports/components/time-entries-table/time-entries-table.component.ts
```diff
@@ -91,4 +85,13 @@ export class TimeEntriesTableComponent implements OnInit, OnDestroy, AfterViewIn
...
+    return this.isURL(uri) ? window.open(uri, '_blank') : '';
+  }
+
+  isURL(uri: string) {
+    const regex = new RegExp('http*', 'g');
+    return regex.test(uri) ? true : false;
```
Angeluz-07, line 95: you could return `regex.test(uri)` directly, as it already returns true or false.

src/app/modules/reports/components/time-entries-table/time-entries-table.component.spec.ts
Angeluz-07, line 92: Let's change uriExpected by `url` in this test also.

src/app/modules/reports/components/time-entries-table/time-entries-table.component.ts
Angeluz-07: Also, a type in the function may help. `isURL(uri: string) : boolean`

thegreatyamori: Yes, the type of the function has not yet been placed.

thegreatyamori:
Good work, only remains to correct the type of the `isURL` function.

src/app/modules/reports/components/time-entries-table/time-entries-table.component.ts
thegreatyamori: Nice ! 🌟

thegreatyamori (approved):

Angeluz-07 (approved):
Good job!.

scastillo-jp (approved):

PR #650, opened 2021-03-19
TT 155 consume user info

thegreatyamori:


debug.log
```diff
@@ -0,0 +1,8 @@
+[1207/120956.471:ERROR:directory_reader_win.cc(43)] FindFirstFile: El sistema no puede encontrar la ruta especificada. (0x3)
```
Angeluz-07, line 1: Remove this file.

src/app/app.module.ts
```diff
@@ -161,9 +162,10 @@ const maskConfig: Partial<IConfig> = {
       TechnologyEffects,
       ProjectTypeEffects,
       EntryEffects,
+      UsersEffects,
```
scastillo-jp, line 165: You need to keep these two imports? `UsersEffects`  `UserEffects`

src/app/modules/shared/components/user/user.component.ts
```diff
@@ -1,6 +1,7 @@
 import { Component, OnInit } from '@angular/core';
 import { AzureAdB2CService } from '../../../login/services/azure.ad.b2c.service';
 
+
```
scastillo-jp, line 4: delete this line

src/app/modules/user/services/user.service.ts
```diff
@@ -0,0 +1,21 @@
+import { Injectable } from '@angular/core';
+import { HttpClient } from '@angular/common/http';
+
```
scastillo-jp, line 3: delete this line

src/app/modules/user/services/user.service.ts
```diff
@@ -0,0 +1,21 @@
...
+  providedIn: 'root',
+})
+export class UserService {
+
+  constructor(private http: HttpClient) {
+  }
+
+  baseUrl = `${environment.timeTrackerApiUrl}/users`;
+
+  loadUser(userId): Observable<any> {
```
scastillo-jp, line 17: I think can change `Observable<any>`  to `Observable<string>`

src/app/app.module.ts
thegreatyamori: Yes, both effects do different things.
```
import { UserEffects as UsersEffects } from './modules/users/store/user.effects';
import { UserEffects } from './modules/users/users/store/user.effects';
```
We found that there is a conflict with the names, especially in the `UserEffects` of users.

Should we change the name?

src/app/modules/user/services/user.service.ts
thegreatyamori: thank you for your observation.
Actually an observable of type `User` is returned.
I willl change it !
🏁

scastillo-jp (approved):

kellycastrof (commented):
A question just to understand, why is there a UserEffect and a UsersEffect that come from the same module?
https://github.com/ioet/time-tracker-ui/blob/872952eee59440b910a96d60cd26cdd1ccaf7339/src/app/app.module.ts#L60-L61

thegreatyamori:
> A question just to understand, why is there a UserEffect and a UsersEffect that come from the same module?
> https://github.com/ioet/time-tracker-ui/blob/872952eee59440b910a96d60cd26cdd1ccaf7339/src/app/app.module.ts#L60-L61
Because there are two nrgx data stores.
The one we made is called `UserEffect` and it is used to query the logged user data.
The first one is `UsersEffects` and it is used to fetch all the registered users. With @Angeluz-07 we decided to add an alias to avoid having to modify the rest of the code since it did not correspond to us.
However, the name should be changed to `UsersEffect` to avoid future conflicts.

LEON12699 (approved):

PR #653, opened 2021-03-24
Tt 188 add and remove users from groups

thegreatyamori:


Angeluz-07 (commented):
Very good job. I think we would need to update the references in all the files:
`add group to user` to `add user to group` and
`remove group to user` to `remove user from group`.

thegreatyamori:
> Very good job. I think we would need to update the references in all the files:
> `add group to user` to `add user to group` and
> `remove group to user` to `remove user from group`.

ok, I'm going to change the references

Angeluz-07 (approved):

scastillo-jp (approved):

PR #658, opened 2021-04-05
feat: TT-190  user user-info service to check if user is admin in UI

thegreatyamori:
- Added ngrx-store-localstorage to persist specific store (user)
- Added Feature Toggle *switch-group* to mantain the old logic
- Added the corresponding tests


src/environments/environment.ts
```diff
@@ -22,6 +22,9 @@ export const GROUPS = {
   ADMIN: 'time-tracker-admin',
   TESTER: 'time-tracker-tester',
 };
+export const TOGGLES = {
+  SWITCH_GROUP: 'switch-group',
```
scastillo-jp, line 26: I have a question:
The functionality of this var is the storage of the state of the switch group?
- if this value changes this value is updated?

src/app/modules/home/home.component.ts
```diff
@@ -1,15 +1,34 @@
...
 
 @Component({
   selector: 'app-home',
   templateUrl: './home.component.html',
-  styleUrls: ['./home.component.scss']
+  styleUrls: ['./home.component.scss'],
 })
-export class HomeComponent implements OnInit {
+export class HomeComponent implements OnInit, OnDestroy {
+  ftSwitchGroup$: Subscription;
```
scastillo-jp, line 14: I prefer use FTSwitchGroup

src/environments/environment.ts
thegreatyamori: No, it just the name of FT. I was using because it was easier to change the name if it necesary and in case you needed to use it in more parts of the application. The value isn't update if FT changes

src/app/modules/home/home.component.ts
thegreatyamori: Okay, I change it !

scastillo-jp (approved):

PR #660, opened 2021-04-09
fix: TT-23 Clear form when adding an entry (Time Entries)

LEON12699:
Add form reset icon in (timeEntries)  
iconoReset

### DIFF
--- a/src/app/modules/shared/components/details-fields/details-fields.component.ts
+++ b/src/app/modules/shared/components/details-fields/details-fields.component.ts
@@ -172,7 +172,7 @@ export class DetailsFieldsComponent implements OnChanges, OnInit {
     });
   }
 
-  cleanFormWithSkipProject(): void {
+  cleanFieldsForm(): void {
     this.cleanForm(true);
   }
 
--- a/src/app/modules/shared/components/details-fields/details-fields.component.ts
+++ b/src/app/modules/shared/components/details-fields/details-fields.component.ts
@@ -19,6 +19,7 @@ import { SaveEntryEvent } from './save-entry-event';
 import { ProjectSelectedEvent } from './project-selected-event';
 import { get } from 'lodash';
 import { DATE_FORMAT } from 'src/environments/environment';
+import { TechnologiesComponent } from '../technologies/technologies.component';
 
 type Merged = TechnologyState & ProjectState & ActivityState & EntryState;
 @Component({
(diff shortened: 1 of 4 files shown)


src/app/modules/shared/components/details-fields/details-fields.component.ts
```diff
@@ -151,11 +153,14 @@ export class DetailsFieldsComponent implements OnChanges, OnInit {
     }
   }
 
-  cleanForm() {
+  cleanForm(skipProject: boolean = false): void {
     this.selectedTechnologies = [];
```
scastillo-jp, line 157: Is necessary add this line  **this.technologies.query = '';** ?

I think with this line you can reset the technologies   **this.selectedTechnologies = [];**

src/app/modules/shared/components/details-fields/details-fields.component.ts
```diff
@@ -167,6 +172,10 @@ export class DetailsFieldsComponent implements OnChanges, OnInit {
     });
   }
 
+  cleanFormWithSkipProject(): void {
```
scastillo-jp, line 175: rename this function to **cleanFieldsForm**.

src/app/modules/time-entries/pages/time-entries.component.html
```diff
@@ -64,13 +64,17 @@
     <div class="modal-content" cdkDrag (cdkDragEnded)="resetDraggablePosition($event)">
       <div class="modal-header" cdkDragHandle>
         <h5 class="modal-title">{{ entryId ? 'Edit Entry' : 'New Entry' }}</h5>
+        <button type="button" class="btn" (click)="detailsFields.cleanFormWithSkipProject()">
```
scastillo-jp, line 67: How work with  #Id **detailsFields**?
I think this should be this.cleanFormWithSkipProject()

src/app/modules/shared/components/details-fields/details-fields.component.ts
LEON12699: With the part of this.`selectedTechnologies =[]`; the tags that appear below the input are cleaned, but the `input` of technologies is not cleaned if there is something in it, that is why it is necessary to add this line, I think it is because this part is another component `app-technologies`  and is not directly associated with the form.

src/app/modules/time-entries/pages/time-entries.component.html
LEON12699: With the `#detailFields` a local template variable associated with the ` app-details-fields ` component is created, this allows access to the functions and data of the ` app-details-fields ` component (child component) in this case when you click on the icon, the function `cleanFormWithSkipProject()` that is in the `details-fields.component.ts` is executed as it is not a function of the component `time-entries.component.ts` can not be accessed with this, but otherwise `this.detailsFields.cleanFormWithSkipProject() ` can be used but i think its not neccesary.

scastillo-jp (approved):

PR #663, opened 2021-04-13
feat: TT-217 Add FT in EntryFieldsComponent component and Add flag at validate current entry and last entry

VanessaIniguez:
Add FT in EntryFieldsComponent component to get  the new changes, add the flag to validate the current entry and last entry, add tests of the Feature Toggle - update-entries.

### DIFF
--- a/src/app/modules/time-clock/components/entry-fields/entry-fields.component.spec.ts
+++ b/src/app/modules/time-clock/components/entry-fields/entry-fields.component.spec.ts
@@ -439,11 +439,8 @@ describe('EntryFieldsComponent', () => {
       update_last_entry_if_overlap: true
     };
     component.newData = mockEntry;
-
     component.isFeatureToggleActivated();
-
     const expected = { update_last_entry_if_overlap: true };
-
     expect(component.newData.update_last_entry_if_overlap).toEqual(expected.update_last_entry_if_overlap);
   });
 
@@ -456,11 +453,8 @@ describe('EntryFieldsComponent', () => {
       update_last_entry_if_overlap: false
     };
     component.newData = mockEntry;
-
     component.isFeatureToggleActivated();
-
     const expected = { update_last_entry_if_overlap: false };
-
     expect(component.newData.update_last_entry_if_overlap).toEqual(expected.update_last_entry_if_overlap);
   });
 });
--- a/src/app/modules/time-clock/components/entry-fields/entry-fields.component.spec.ts
+++ b/src/app/modules/time-clock/components/entry-fields/entry-fields.component.spec.ts
@@ -469,13 +469,6 @@ describe('EntryFieldsComponent', () => {
       isFeatureToggleActivated.subscribe((value) => expect(value).toEqual(toggleValue));
     });
   });
-
-  // fit('should return the id of the active user', () => {
-
-  //   const userId = 'user_id';
-
-  //   expect( component.getOwnerId()).
-  // });
 });
 
 
--- a/src/app/modules/time-clock/components/entry-fields/entry-fields.component.spec.ts
+++ b/src/app/modules/time-clock/components/entry-fields/entry-fields.component.spec.ts
@@ -1,4 +1,4 @@
-import { Subscription, of } from 'rxjs';
+import { Subscription, of, Observable } from 'rxjs';
 import { LoadActiveEntry, EntryActionTypes, UpdateEntry } from './../../store/entry.actions';
 import { ActivityManagementActionTypes } from './../../../activities-management/store/activity-management.actions';
 import {waitForAsync, ComponentFixture, TestBed} from '@angular/core/testing';
@@ -430,7 +430,7 @@ describe('EntryFieldsComponent', () => {
     expect(component.actionSetDateSubscription.unsubscribe).toHaveBeenCalled();
   });
 
-  it('The flag "update_last_entry_if_overlap" is added to the "newData" when feature flag "update-entries" is enabled for user', () => {
+  it('when feature-toggle "update-entries" enable for the user, the updateEntry function is executes to update the entries', () => {
     spyOn(featureManagerService, 'isToggleEnabledForUser').and.returnValue(of(true));
 
     const mockEntry = { ...entry,
@@ -444,7 +444,7 @@ describe('EntryFieldsComponent', () => {
     expect(component.newData.update_last_entry_if_overlap).toEqual(expected.update_last_entry_if_overlap);
   });
 
-  it('The flag "update_last_entry_if_overlap" is not added to the "newData" when feature flag "update-entries" is disable for user', () => {
+  it('when FT "update-entries" disable for the user,the UpdateCurrentOrLastEntry function is called to update the entries', () => {
     spyOn(featureManagerService, 'isToggleEnabledForUser').and.returnValue(of(false));
 
     const mockEntry = { ...entry,
@@ -457,4 +457,25 @@ describe('EntryFieldsComponent', () => {
     const expected = { update_last_entry_if_overlap: false };
     expect(component.newData.update_last_entry_if_overlap).toEqual(expected.update_last_entry_if_overlap);
   });
+
+  const toggleValues = [true, false];
+  toggleValues.map((toggleValue) => {
+    it(`when FeatureToggle is ${toggleValue} should return ${toggleValue}`, () => {
+      spyOn(featureManagerService, 'isToggleEnabledForUser').and.returnValue(of(toggleValue));
+
+      const isFeatureToggleActivated: Observable<boolean> = component.isFeatureToggleActivated();
+
+      expect(featureManagerService.isToggleEnabledForUser).toHaveBeenCalled();
+      isFeatureToggleActivated.subscribe((value) => expect(value).toEqual(toggleValue));
+    });
+  });
+
+  // fit('should return the id of the active user', () => {
+
+  //   const userId = 'user_id';
+
+  //   expect( component.getOwnerId()).
+  // });
 });
+
+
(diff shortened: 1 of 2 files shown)


src/app/modules/time-clock/components/entry-fields/entry-fields.component.ts
```diff
@@ -15,7 +15,11 @@ import { ToastrService } from 'ngx-toastr';
 import { formatDate } from '@angular/common';
 import { getTimeEntriesDataSource } from '../../store/entry.selectors';
 import { DATE_FORMAT } from 'src/environments/environment';
-import { Subscription } from 'rxjs';
+import { Subscription, Observable } from 'rxjs';
+
```
scastillo-jp, line 19: delete this white line

src/app/modules/time-clock/components/entry-fields/entry-fields.component.spec.ts
```diff
@@ -425,4 +429,38 @@ describe('EntryFieldsComponent', () => {
...
+  it('The flag "update_last_entry_if_overlap" is not added to the "newData" when feature flag "update-entries" is disable for user', () => {
+    spyOn(featureManagerService, 'isToggleEnabledForUser').and.returnValue(of(false));
+
+    const mockEntry = { ...entry,
+      start_date : moment().format(DATE_FORMAT_YEAR),
+      start_hour : moment().format('HH:mm'),
+      update_last_entry_if_overlap: false
+    };
+    component.newData = mockEntry;
+
```
jeffqev, line 459: delete this white line

src/app/modules/time-clock/components/entry-fields/entry-fields.component.ts
```diff
@@ -145,10 +157,20 @@ export class EntryFieldsComponent implements OnInit, OnDestroy {
...
+      this.store.dispatch(new entryActions.UpdateEntry({ ...this.newData, ...this.entryForm.value }));
+    } else {
+      this.store.dispatch(new entryActions.UpdateCurrentOrLastEntry({ ...this.newData, ...this.entryForm.value }));
+    }
     this.showTimeInbuttons = false;
   }
 
+  getOwnerId(){
+    return this.azureService.getUserId();
+  }
```
scastillo-jp, line 172: @VanessaIniguez is necessary add a test to cover this method.

src/app/modules/time-clock/components/entry-fields/entry-fields.component.ts
```diff
@@ -178,4 +200,9 @@ export class EntryFieldsComponent implements OnInit, OnDestroy {
     this.loadActiveEntrySubscription.unsubscribe();
     this.actionSetDateSubscription.unsubscribe();
   }
+
+  isFeatureToggleActivated(): Observable<boolean> {
+    return this.featureManagerService.isToggleEnabledForUser('update-entries')
+      .pipe(map((enabled: boolean) => enabled));
+  }
```
scastillo-jp, line 207: Is necessary add a  test when the feature toggle is enable o disable.

you can use this test

```
  const toggleValues = [true, false];
  toggleValues.map((toggleValue) => {
    it(`when FeatureToggle is ${toggleValue} should return ${toggleValue}`, () => {
      spyOn(service, 'isToggleEnabledForUser').and.returnValue(of(toggleValue));

      const isFeatureToggleActivated: Observable<boolean> = component.isFeatureToggleActivated();

      expect(service.isToggleEnabledForUser).toHaveBeenCalled();
      isFeatureToggleActivated.subscribe((value) => expect(value).toEqual(toggleValue));
    });
  });
```

src/app/modules/time-clock/components/entry-fields/entry-fields.component.spec.ts
```diff
@@ -425,4 +429,32 @@ describe('EntryFieldsComponent', () => {
     expect(component.loadActiveEntrySubscription.unsubscribe).toHaveBeenCalled();
     expect(component.actionSetDateSubscription.unsubscribe).toHaveBeenCalled();
   });
+
+  it('The flag "update_last_entry_if_overlap" is added to the "newData" when feature flag "update-entries" is enabled for user', () => {
```
kellycastrof, line 433: Maybe this could be more descriptive: when feature ... enable for the user and running entry update is called

src/app/modules/time-clock/components/entry-fields/entry-fields.component.spec.ts
```diff
@@ -425,4 +429,32 @@ describe('EntryFieldsComponent', () => {
...
+      start_hour : moment().format('HH:mm'),
+      update_last_entry_if_overlap: true
+    };
+    component.newData = mockEntry;
+    component.isFeatureToggleActivated();
+    const expected = { update_last_entry_if_overlap: true };
+    expect(component.newData.update_last_entry_if_overlap).toEqual(expected.update_last_entry_if_overlap);
+  });
+
+  it('The flag "update_last_entry_if_overlap" is not added to the "newData" when feature flag "update-entries" is disable for user', () => {
```
kellycastrof, line 447: The same for this test

src/app/modules/time-clock/components/entry-fields/entry-fields.component.ts
VanessaIniguez: Thanks Sandro, I now implement it

src/app/modules/time-clock/components/entry-fields/entry-fields.component.spec.ts
VanessaIniguez: Thanks Kelly for your comments, I will consider what you say

src/app/modules/time-clock/components/entry-fields/entry-fields.component.ts
```diff
@@ -145,10 +157,20 @@ export class EntryFieldsComponent implements OnInit, OnDestroy {
       return;
     }
     this.entryForm.patchValue({ start_date: newHourEntered });
-    this.store.dispatch(new entryActions.UpdateCurrentOrLastEntry({ ...this.newData, ...this.entryForm.value }));
+    if (this.isTestUser) {
+      this.newData.owner_id = this.getOwnerId();
```
Angeluz-07, line 161: Mmm I am thinking if you really need the owner_id. It should be already present in the data.

src/app/modules/time-clock/components/entry-fields/entry-fields.component.ts
```diff
@@ -145,10 +157,20 @@ export class EntryFieldsComponent implements OnInit, OnDestroy {
...
+      this.newData.update_last_entry_if_overlap = true;
```
Angeluz-07, line 162: I think this variable should not be here. There is an action called, update entry running. Try putting this assigment there.

src/app/modules/time-clock/components/entry-fields/entry-fields.component.ts
```diff
@@ -145,10 +157,20 @@ export class EntryFieldsComponent implements OnInit, OnDestroy {
...
+      this.store.dispatch(new entryActions.UpdateEntry({ ...this.newData, ...this.entryForm.value }));
```
Angeluz-07, line 163: I think the action to be called is updateEntryRunning. `    this.store.dispatch(new entryActions.UpdateEntryRunning({ ...this.newData, ...this.entryForm.value }));
`

src/app/modules/time-clock/components/entry-fields/entry-fields.component.ts
Angeluz-07: You may realize that when you update an entry, you dont need to set the owner_id.

src/app/modules/time-clock/components/entry-fields/entry-fields.component.ts
```diff
@@ -145,10 +157,20 @@ export class EntryFieldsComponent implements OnInit, OnDestroy {
       return;
     }
     this.entryForm.patchValue({ start_date: newHourEntered });
-    this.store.dispatch(new entryActions.UpdateCurrentOrLastEntry({ ...this.newData, ...this.entryForm.value }));
+    if (this.isTestUser) {
```
Angeluz-07, line 160: This is not critical, but consider that this flag is not about if the user is tester or not. This flag is about if FT is enabled or not.

Angeluz-07 (commented):
Good Job. Only some comments that can make your solution cleaner. Also dont forget after validating these changes. All the code around UpdateCurrentOrLastEntry should be removed. -> `https://github.com/ioet/time-tracker-ui/search?q=updateCurrentOrLastEntry`

src/app/modules/time-clock/components/entry-fields/entry-fields.component.ts
VanessaIniguez: Thanks for the comments, I will give the flag another name

src/app/modules/time-clock/components/entry-fields/entry-fields.component.ts
VanessaIniguez: Hi Roberto, thanks for your observations, the reason for sending the owner_id is because in the backend this property is used to query database. And when I tried connecting locally the backend didn't return anything because the property was not sent. Now with updateWntryRunning method I will see if I don't need to send this property. Thanks Roberto.

src/app/modules/time-clock/components/entry-fields/entry-fields.component.ts
thegreatyamori, line 207: I think this method is easier to delete if it's put into a service. Since you can decouple the logic from the component and when the FT is no longer used you should just remove the service and the references in the component.
What do you think @scastillo-jp ?

thegreatyamori (changes requested):

src/app/modules/time-clock/components/entry-fields/entry-fields.component.ts
thegreatyamori: Oh, It's easier to delete in the test part too.

thegreatyamori (commented):
Good job !

scastillo-jp (approved):

PR #671, opened 2021-04-26
fix: TT-230 implement datepicker safari on Reports

LEON12699:
datepicker on reports

### DIFF
--- a/src/app/modules/reports/components/time-range-form/time-range-form.component.ts
+++ b/src/app/modules/reports/components/time-range-form/time-range-form.component.ts
@@ -29,8 +29,8 @@ export class TimeRangeFormComponent implements OnInit {
 
   setInitialDataOnScreen() {
     this.reportForm.setValue({
-      startDate: formatDate(moment().startOf('week').toString(), DATE_FORMAT, 'en'),
-      endDate: formatDate(moment().endOf('week').toString(), DATE_FORMAT, 'en')
+      startDate: formatDate(moment().startOf('week').format('l'), DATE_FORMAT, 'en'),
+      endDate: formatDate(moment().endOf('week').format('l'), DATE_FORMAT, 'en')
     });
     this.onSubmit();
   }
--- a/src/app/modules/shared/components/input-date/input-date.component.html
+++ b/src/app/modules/shared/components/input-date/input-date.component.html
@@ -1,7 +1,7 @@
 <input
   [value]="value"
   [disabled]="isDisabled"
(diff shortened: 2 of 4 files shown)


src/app/modules/reports/components/time-range-form/time-range-form.component.ts
```diff
@@ -29,8 +29,8 @@ export class TimeRangeFormComponent implements OnInit {
 
   setInitialDataOnScreen() {
     this.reportForm.setValue({
-      startDate: formatDate(moment().startOf('week').toString(), DATE_FORMAT, 'en'),
-      endDate: formatDate(moment().endOf('week').toString(), DATE_FORMAT, 'en')
+      startDate: formatDate(moment().startOf('week').format('l'), DATE_FORMAT, 'en'),
```
thegreatyamori, line 32: I have a question
Why you use a localized formats `.format('l')`?

src/app/modules/shared/components/input-date/input-date.component.html
```diff
@@ -1,10 +1,12 @@
 <input
   [value]="value"
   [disabled]="isDisabled"
-  (input)="onInput($event.target.value)"
-  type="date"
+  (dateInput)="onInput($event.target.value.format('YYYY-MM-DD'))"
```
thegreatyamori, line 4: I think it is a better way, to apply the format in component.ts and use DATE_FORMAT

src/app/modules/reports/components/time-range-form/time-range-form.component.ts
LEON12699: If this is not used you get an error, the mat-datepicker component uses MatMomentDateModule which formats the dates inside the datepicker by default using the `format 'l'`, making use of the moment library to return the dates. The moment objects are mutable while by default ngrx 10 use strict immutability and freeze objects, it seems to be a problem of the MatMomentDateModule library, anyway in these links you can find information about errors that I have based on and if there is a better way I am open to suggestions:

- https://github.com/datorama/akita/issues/124
- https://github.com/moment/momentjs.com/issues/292
- https://github.com/ngrx/platform/issues/2690

It can be used only in the  ngOnInit() of the component, but I think it can be better prevented in this way,

thegreatyamori: I understand. For the time being you are forced to keep the code like this.

thegreatyamori (commented):
Good job, just check what's going on with the CI to make sure everything is correct.

scastillo-jp (approved):

PR #674, opened 2021-04-30
fix: TT-226 Show user an error message when they try to save a time e…

eduardisrael:
[TT-226](https://ioetec.atlassian.net/browse/TT-226)

This PR implement 'When the user tries to save, an error message should be shown that the clock out time can’t be before the clock in time'.

### DIFF
--- a/src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
+++ b/src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
@@ -576,6 +576,12 @@ describe('DetailsFieldsComponent', () => {
     const foundActivity = component.findInactiveActivity(state.activities.data);
     expect(foundActivity).toEqual(expectedActivity);
   });
+
+  it('should return false when the start time entry is not greater than the end time', () => {
+    const result = component.isStartTimeEntryAfterEndedEntry();
+
+    expect(result).toBeFalse();
+  });
   /*
    TODO As part of https://github.com/ioet/time-tracker-ui/issues/424 a new parameter was added to the details-field-component,
    and now these couple of tests are failing. A solution to this error might be generate a Test Wrapper Component. More details here:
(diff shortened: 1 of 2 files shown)


src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
```diff
@@ -576,6 +576,12 @@ describe('DetailsFieldsComponent', () => {
     const foundActivity = component.findInactiveActivity(state.activities.data);
     expect(foundActivity).toEqual(expectedActivity);
   });
+
+  it('should return false when the start time entry is not greater than the end time', () => {
+    const result = component.isStartTimeEntryAfterEndedEntry();
+
+    expect(result).toBeFalse();
+  });
```
thegreatyamori, line 584: In this part I think it we need more test cases to ensure that the output will not fail. I've prepared a table with the all combinations and outputs.

I would suggest to select the most relevant cases. 😄 

image

thegreatyamori (commented):
Good Job !!

src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
LEON12699: For what @thegreatyamori  recommends you might consider using ` parameterized tests` similar to those found in line 538.
You could pass the values to the form using the patchValue something like this:

```
const FormDateValues = {
      start_date: '2021-05-15',
      end_date: '2021-04-14',
      start_hour: '15:05',
      end_hour: '17:00',
  };

    component.entryForm.patchValue(FormDateValues);
    const result = component.isStartTimeEntryAfterEndedEntry();
```

LEON12699:
LGTM 🚀 !

thegreatyamori (approved):

LEON12699 (approved):

PR #679, opened 2021-05-04
Tt 218 dont allow deleting projects

jeffqev:
Projects can be archived or activated

Don't allow deleting projects

### DIFF
--- a/src/app/modules/customer-management/components/projects/components/project-list/project-list.component.ts
+++ b/src/app/modules/customer-management/components/projects/components/project-list/project-list.component.ts
@@ -2,12 +2,10 @@ import { Component, OnInit, OnDestroy } from '@angular/core';
 import { Store, select } from '@ngrx/store';
 import { Subscription } from 'rxjs';
 import { ITEMS_PER_PAGE } from 'src/environments/environment';
-import { Project } from 'src/app/modules/shared/models';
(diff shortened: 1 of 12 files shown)


src/app/modules/customer-management/components/projects/components/project-list/project-list.component.ts
```diff
@@ -47,9 +69,22 @@ export class ProjectListComponent implements OnInit, OnDestroy {
...
+  switchStatus(item: ProjectUI): void {
+    if (item.key !== 'inactive') {
+      this.openModal(item);
+    } else {
+      this.showModal = false;
+      this.store.dispatch(new UnarchiveProject(item.id));
+    }
+  }
+
+  setActive(status: any): string {
```
thegreatyamori, line 87: I think this is a redundant method.
the status variable  can return `'inactive'` | `'active'` or only `'inactive'`?

jeffqev: In some cases the endpoint return `active or null` for active items so there I make sure just return `active or inactive`

thegreatyamori: Ohh, okay. I understand it !

scastillo-jp (approved):

PR #681, opened 2021-05-06
TT-236 When you add a new entry and check "I'm working on this" you should take the current time for "Time in"

thegreatyamori:
We found a bug when changing the value of the `goingToWorkOnThis` variable in `ngOnChanges`, it was setting a value to undefined because the object being checked did not have the property at the time of adding a new time entry. We fixed the error using "Nullish coalescing operator" in addition to correcting the bug mentioned in the ticket.


jeffqev (approved):
Good job, It works fine

PR #700, opened 2021-06-28
fix: TT-282 added conditions in method onSubmit and added array of activities in component project-list-hover

kevinjlope:
### Problem
If somebody added a new time entry and if in the DB not existing register about of the information the attribute activity_id. The application can not show the information about time entry and the reports, and return: the data can not be loaded. 

### The solution
When the user selects a new project, in the panel time clock, to default send an activity_id, and the DB.

### DIFF
--- a/src/app/modules/time-clock/components/entry-fields/entry-fields.component.html
+++ b/src/app/modules/time-clock/components/entry-fields/entry-fields.component.html
@@ -6,8 +6,9 @@
       class="form-control"
       formControlName="activity_id"
       [class.is-invalid]="activity_id.invalid && activity_id.touched"
+      #autofocus      
       required>
-      <option *ngFor="let activity of activities" value="{{ activity.id }}">{{ activity.name }}</option>
+      <option *ngFor="let activity of activities" value="{{ activity.id }}" class = "id">{{ activity.name }}</option>
     </select>
   </div>
 
--- a/src/app/modules/time-clock/components/project-list-hover/project-list-hover.component.ts
+++ b/src/app/modules/time-clock/components/project-list-hover/project-list-hover.component.ts
@@ -14,6 +14,11 @@ import {
 } from './../../../customer-management/components/projects/components/store/project.selectors';
 import { EntryActionTypes } from './../../store/entry.actions';
 import { getActiveTimeEntry } from './../../store/entry.selectors';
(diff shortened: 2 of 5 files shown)


src/app/modules/time-clock/components/entry-fields/entry-fields.component.html
```diff
@@ -7,6 +7,7 @@
       formControlName="activity_id"
       [class.is-invalid]="activity_id.invalid && activity_id.touched"
       required>
+      <option value="" selected="selected"></option>
```
jcalarcon98, line 10: I think this change is causing this behavior at the time of displaying the activities:
image
I think from my point of view an empty option should not be displayed.

jcalarcon98 (commented):
Good Job Kevin, just a couple of things to say:
The main bug is solved with this PR, however, I think that the transition of putting a default activity is not transparent, for example, once you do Clock in and if by default you assign the first activity of the list of activities, the element select activities can take the focus, it is something that Fanny mentioned in the review of this bug, analyze if it is possible or it is not too complicated to make that behavior so that it can be implemented. 
And please check my comments.

jcalarcon98:
> Good Job Kevin  , just a couple of things to say:
> The main bug is solved with this PR, however, I think that the transition of putting a default activity is not transparent, for example, once you do Clock in and if by default you assign the first activity of the list of activities, the element select activities can take the focus, it is something that Fanny mentioned in the review of this bug, analyze if it is possible or it is not too complicated to make that behavior so that it can be implemented.
> And please check my comments in the code.

src/app/modules/time-clock/components/project-list-hover/project-list-hover.component.ts
```diff
@@ -88,6 +99,7 @@ export class ProjectListHoverComponent implements OnInit, OnDestroy {
       start_date: new Date().toISOString(),
       timezone_offset: new Date().getTimezoneOffset(),
       technologies: [],
+      activity_id: this.activities[0].id,
```
scastillo-jp, line 102: In this index you can use 
`import { head } from 'lodash';`
`activity_id: _.head(this.activities.id)`  OR `activity_id:  _.head(this.activities).id`

Try with this options.

scastillo-jp (approved):

PR #701, opened 2021-07-06
fix: TT-283 added try catch to handle method rerenderDataTable

kevinjlope:
### Problem
Cannot read property 'clientWidth', this problem present in the section of Reports.

### Solution
Added a try-catch in the method rerenderDataTable. This method is used to handle the dimensions of the reports' table.

### DIFF
--- a/src/app/modules/reports/components/time-entries-table/time-entries-table.component.ts
+++ b/src/app/modules/reports/components/time-entries-table/time-entries-table.component.ts
@@ -2,7 +2,6 @@ import { formatDate } from '@angular/common';
 import { AfterViewInit, Component, OnDestroy, OnInit, ViewChild} from '@angular/core';
 import { select, Store } from '@ngrx/store';
 import { DataTableDirective } from 'angular-datatables';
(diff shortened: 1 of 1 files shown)


src/app/modules/reports/components/time-entries-table/time-entries-table.component.ts
```diff
@@ -82,11 +82,15 @@ export class TimeEntriesTableComponent implements OnInit, OnDestroy, AfterViewIn
   private rerenderDataTable(): void {
     if (this.dtElement && this.dtElement.dtInstance) {
       this.dtElement.dtInstance.then((dtInstance: DataTables.Api) => {
```
scastillo-jp, line 84: I think it is not a correct way to implement a try catch, you can use the if / else or else try / catch but not the two together.

I am attaching a link with information on the correct use of a try / catch.
https://javascript.plainenglish.io/the-dark-side-of-typescript-try-catch-deeded18ba0d

scastillo-jp (approved):

PR #708, opened 2021-07-23
Tt 289 cannot edit the date of an entry

DrFreud1:
An additional comparison was added in order to allow the user to change the date of an entry. Because of that, certain tests were removed because they were no longer relevant. However, the method that was modified has other tests relevant for what it actually does.

To summarize, one method was modified, two tests were removed because no longer needed.

### DIFF
--- a/src/app/modules/shared/components/details-fields/details-fields.component.ts
+++ b/src/app/modules/shared/components/details-fields/details-fields.component.ts
@@ -269,7 +269,6 @@ export class DetailsFieldsComponent implements OnChanges, OnInit {
     const updatedHour = this.entryForm.value[hour];
     const updatedDate = new Date(`${entryFormDate}T${updatedHour.trim()}`).toISOString();
     const initialDate = get(this.entryToEdit, date, updatedDate);
-    const initialHour = formatDate(get(this.entryToEdit, date, updatedDate), 'HH:mm', 'en');
     const dateHasNotChanged = (initialDate === updatedDate);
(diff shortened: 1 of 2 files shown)


kevinjlope (approved):

src/app/modules/shared/components/details-fields/details-fields.component.ts
```diff
@@ -270,7 +270,7 @@ export class DetailsFieldsComponent implements OnChanges, OnInit {
     const updatedDate = new Date(`${entryFormDate}T${updatedHour.trim()}`).toISOString();
     const initialDate = get(this.entryToEdit, date, updatedDate);
     const initialHour = formatDate(get(this.entryToEdit, date, updatedDate), 'HH:mm', 'en');
-    const dateHasNotChanged = updatedHour === initialHour;
+    const dateHasNotChanged = (initialDate === updatedDate) && (updatedHour === initialHour);
```
jcalarcon98, line 273: Is the second condition necessary? For example, I deleted the second condition because it seems to me that it never gets to that point, and everything works correctly.

jcalarcon98 (commented):
Good Job 💥 Santiago, please check my comments and if you consider them necessary you can apply them.

src/app/modules/shared/components/details-fields/details-fields.component.ts
diegotony: @jase156

PR #710, opened 2021-07-28
fix: TT-299 modified state machine

DrFreud1:
When archiving/activating an activity, the data of that entry is shown on the interface which is not desirable. To fix the problem, certain variables were changed in the state machine of angular.


jcalarcon98 (commented):
Great Job @DrFreud1 💥.

I have reviewed the changes made and I think the following property: 
```ts 
activityIdToEdit: undefined
```
If it has a value of undefined, it is technically no longer required, so it would be best to remove it. 

I also noticed that some comments have been left in the code, these should not be added. Please review my suggestions and if you feel it is necessary you can implement it.

PR #713, opened 2021-08-04
Tt 299 activities when active inactive an activity not display information in name description

DrFreud1:
### Problem

In the UI, when archiving or unarchiving an activity, all the content of it was shown on the UI, as shown in the picture below. This was not visually comfortable for the user, and thus the content does not need to appear when performing the actions mentioned at the beginning.

# Solution

The variables which were causing this visual problem were removed and therefore, the tests needed for those variables are no longer necessary and were also removed.


jcalarcon98 (approved):
Good job @DrFreud1  🥇

ridouku (approved):

scastillo-jp (approved):

PR #716, opened 2021-08-06
fix: TT-304 Handle message: the data could not be load

edgardavid2015:
**Problem**
------------
In the time-entries section, when there is no data, the message _"The data could not be load"_ is displayed.
Captura de Pantalla 2021-08-06 a la(s) 12 27 58

**Solution**
------------
When there is no data, the message _"The data could not be load"_ must not be displayed, instead the message _"No data available in table"_ should be displayed in the table section. That is the purpose of this pull request.
Captura de Pantalla 2021-08-06 a la(s) 12 19 29

### DIFF
--- a/src/app/modules/time-entries/pages/time-entries.component.ts
+++ b/src/app/modules/time-entries/pages/time-entries.component.ts
@@ -1,7 +1,7 @@
-import { Component, OnDestroy, OnInit } from '@angular/core';
+import { Component, OnDestroy, OnInit, ViewChild, AfterViewInit } from '@angular/core';
 import { ActionsSubject, select, Store } from '@ngrx/store';
 import { ToastrService } from 'ngx-toastr';
-import { Observable, Subscription } from 'rxjs';
+import { Observable, Subscription, Subject } from 'rxjs';
 import { delay, filter } from 'rxjs/operators';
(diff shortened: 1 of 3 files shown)


src/app/modules/time-entries/pages/time-entries.component.ts
```diff
@@ -216,4 +235,15 @@ export class TimeEntriesComponent implements OnInit, OnDestroy {
       });
     }
   }
+
+  private rerenderDataTable(): void {
+    if (this.dtElement && this.dtElement.dtInstance) {
+      this.dtElement.dtInstance.then((dtInstance: DataTables.Api) => {
+          dtInstance.destroy();
+          this.dtTrigger.next();
```
scastillo-jp, line 243: @edgardavid2015  Is necessary coverage this line with an unit test or update the current test

edgardavid2015: I already created the unit test for this lines, thank you so much for the advice and the revision.

DrFreud1 (approved):

jcalarcon98 (approved):
Good Job Edgar 💥

scastillo-jp (approved):

PR #720, opened 2021-08-12
fix: TT-310 Reports do not work with groups fixed

jcalarcon98:
## Description
Currently, the backend verifies the user has the `admin` role to visualize information in the reports section.
However, the Time Tracker UI, in the Users section, only adds the user to the group of `time-tracker-admin` but does not add the role.
## Solution
With this PR each time a user is added to the admin group, the admin role is also added, and if the user is removed, the admin role is removed as well.

### DIFF
--- a/src/app/modules/users/components/users-list/users-list.component.html
+++ b/src/app/modules/users/components/users-list/users-list.component.html
@@ -1,5 +1,9 @@
 <div class="table-responsive">
-  <table *ngIf="users" class="table table-sm table-bordered table-striped mb-0" datatable [dtTrigger]="dtTrigger"
+  <table
+    *ngIf="users"
+    class="table table-sm table-bordered table-striped mb-0"
+    datatable
+    [dtTrigger]="dtTrigger"
     [dtOptions]="dtOptions">
     <thead class="thead-blue">
       <tr class="d-flex flex-wrap">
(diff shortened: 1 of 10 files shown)


edgardavid2015 (approved):
I watched how @jcalarcon98 worked on the development of this solution. Everything looks great to me.

src/app/modules/users/components/users-list/users-list.component.html
```diff
@@ -1,11 +1,6 @@
 <div class="table-responsive">
-  <table
-    *ngIf="users"
-    class="table table-sm table-bordered table-striped mb-0"
-    datatable
-    [dtTrigger]="dtTrigger"
-    [dtOptions]="dtOptions"
-  >
+  <table *ngIf="users" class="table table-sm table-bordered table-striped mb-0" datatable [dtTrigger]="dtTrigger"
```
scastillo-jp, line 2: @jcalarcon98 restore to original format or previous indentation code

src/app/modules/users/components/users-list/users-list.component.ts
```diff
@@ -28,10 +25,11 @@ export class UsersListComponent implements OnInit, OnDestroy, AfterViewInit {
   dtOptions: any = {};
   switchGroupsSubscription: Subscription;
 
-  constructor(
-    private store: Store<User>,
-    private actionsSubject$: ActionsSubject,
-  ) {
+  public get ROLES() {
+    return ROLES;
```
scastillo-jp, line 29: resolve this warning test coverage.

src/app/modules/users/components/users-list/users-list.component.html
jcalarcon98: Solved!

src/app/modules/users/components/users-list/users-list.component.ts
jcalarcon98: Solved!

PR #722, opened 2021-08-15
fix: TT-311 Can not update entries fixed

jcalarcon98:
## Description
Currently there is an error with some entries that does not allow to modify their start or end time, the error that is generated is the following:
image
 
When attempting to modify the following entries:
image

This error occurs because in the current implementation to modify the entries they are not taking into account the seconds of the entries, in other words, only hours and minutes are taken into account.

## Solution
With this PR, this implementation has been fixed and now takes into account all the information of the start/end time of the entry.

### DIFF
--- a/src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
+++ b/src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
@@ -660,13 +660,13 @@ describe('DetailsFieldsComponent', () => {
         date: '2021-04-20',
         hourAndMinutes: '10:00',
         seconds: 20,
-        miliseconds: 10,
+        miliseconds: 100,
       },
       {
         date: '2020-09-21',
         hourAndMinutes: '08:00',
         seconds: 10,
-        miliseconds: 10,
+        miliseconds: 100,
       },
       {
         date: '2021-04-15',
@@ -678,15 +678,15 @@ describe('DetailsFieldsComponent', () => {
         date: '2019-03-29',
         hourAndMinutes: '12:00',
         seconds: 30,
-        miliseconds: 40,
+        miliseconds: 400,
       },
     ];
 
     const expectedISODates = [
-      '2021-04-20T15:00:20.100Z',
-      '2020-09-21T13:00:10.100Z',
-      '2021-04-16T04:00:00.000Z',
-      '2019-03-29T17:00:30.400Z',
+      new Date('2021-04-20T10:00:20.100').toISOString(),
+      new Date('2020-09-21T08:00:10.100').toISOString(),
+      new Date('2021-04-15T23:00:00.000').toISOString(),
+      new Date('2019-03-29T12:00:30.400').toISOString(),
     ];
 
     fakeDates.forEach(({ date, hourAndMinutes, seconds, miliseconds }, fakeDateIndex) => {
--- a/src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
+++ b/src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
@@ -701,7 +701,7 @@ describe('DetailsFieldsComponent', () => {
     const expectedISOFormatNumbers = ['01', '02', '03', '04', '20', '30', '40', '32', '45'];
 
     numbersForTest.forEach((currentNumber, numberIndex) => {
-      const numberinISOFormat = component.getNumberInISOFornat(currentNumber);
+      const numberinISOFormat = component.getNumberInISOFormat(currentNumber);
       expect(numberinISOFormat).toBe(expectedISOFormatNumbers[numberIndex]);
     });
   });
--- a/src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
+++ b/src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
@@ -141,8 +141,8 @@ describe('DetailsFieldsComponent', () => {
       uri: 'ticketUri',
       start_date: '',
       end_date: '',
-      start_hour: '00:00:10',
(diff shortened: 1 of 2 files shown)


scastillo-jp (approved):

src/app/modules/shared/components/details-fields/details-fields.component.ts
```diff
@@ -264,12 +264,37 @@ export class DetailsFieldsComponent implements OnChanges, OnInit {
...
-    const updatedDate = new Date(`${entryFormDate}T${updatedHour.trim()}`).toISOString();
+  getDateISOFormat(date: string, hourAndMinutes: string, seconds, miliseconds: number): string {
+    hourAndMinutes = hourAndMinutes.trim();
+    const secondsInISOFormat: string = this.getNumberInISOFornat(seconds);
+    const milisecondsInISOFormat: string = this.getNumberInISOFornat(miliseconds);
+    const ISOFormat = `${date}T${hourAndMinutes}:${secondsInISOFormat}.${milisecondsInISOFormat}`;
+    return new Date(ISOFormat).toISOString();
+  }
+
+  getNumberInISOFornat(numberToFormat: number): string {
```
ridouku, line 275: Change this, is format no fornat

src/app/modules/shared/components/details-fields/details-fields.component.ts
```diff
@@ -264,12 +264,37 @@ export class DetailsFieldsComponent implements OnChanges, OnInit {
...
+    const limitSingleNumber = 9;
```
ridouku, line 276: Set the types in this constants

src/app/modules/shared/components/details-fields/details-fields.component.ts
```diff
@@ -264,12 +264,37 @@ export class DetailsFieldsComponent implements OnChanges, OnInit {
...
+    const isNumberGreaterThanLimitNumber = numberToFormat > limitSingleNumber;
+    return isNumberGreaterThanLimitNumber ? numberToFormat.toString() : `0${numberToFormat}`;
+  }
+
+  dateToSubmit(date: string, hour: string) {
+    const timeEntryISODate: string = get(this.entryToEdit, date);
+    let seconds = 0;
+    let miliseconds = 0;
```
ridouku, line 284: Set the types in this constants `seconds, miliseconds`

src/app/modules/shared/components/details-fields/details-fields.component.ts
```diff
@@ -264,12 +264,37 @@ export class DetailsFieldsComponent implements OnChanges, OnInit {
...
+      const timeEntryDate: Date = new Date(timeEntryISODate);
+      seconds = timeEntryDate.getSeconds();
+      miliseconds = timeEntryDate.getMilliseconds();
+    }
+
+    const newEntryDate: string = this.entryForm.value[date];
+    const newEntryHour: string = this.entryForm.value[hour];
+
+    const updatedDate: string = this.getDateISOFormat(newEntryDate, newEntryHour, seconds, miliseconds);
     const initialDate = get(this.entryToEdit, date, updatedDate);
```
ridouku, line 296: Missing type

src/app/modules/shared/components/details-fields/details-fields.component.ts
```diff
@@ -264,12 +264,37 @@ export class DetailsFieldsComponent implements OnChanges, OnInit {
...
-    const dateHasNotChanged = (initialDate === updatedDate);
+    const dateHasNotChanged = initialDate === updatedDate;
```
ridouku, line 297: Missing type

src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
```diff
@@ -652,6 +653,58 @@ describe('DetailsFieldsComponent', () => {
     componentToTest.onSubmit();
     expect(toastrServiceStub.error).not.toHaveBeenCalled();
   });
+
+  it('Should return a date in ISO format given a date, hour, minute, second and milisecond', () => {
```
ridouku, line 657: the test is too long, you could refactor this to reduce the size, you could define your constants outside and set the values in the beforeEach method or make a method that sets the values and then call it.

ridouku (changes requested):

src/app/modules/shared/components/details-fields/details-fields.component.ts
jcalarcon98: We have a rule in our linter: Type of number trivially inferred from a literal number.
image
You can check it

src/app/modules/shared/components/details-fields/details-fields.component.ts
jcalarcon98: Same as above

src/app/modules/shared/components/details-fields/details-fields.component.spec.ts
jcalarcon98: I think it is unnecessary to define in the beforeEach because these values are only needed in this Test. You can see the AAA pattern to give you a clearer idea

src/app/modules/shared/components/details-fields/details-fields.component.ts
jcalarcon98: Same as above

src/app/modules/shared/components/details-fields/details-fields.component.ts
jcalarcon98: It was not noticed, I'll solve it right away.

ridouku (approved):

DrFreud1 (approved):

PR #726, opened 2021-08-17
Tt 313 run the unit test before run the pre commit

DrFreud1:
### Observation
It would be better if we run unit tests before commiting, and only commit and push after tests are successful.

### Solution

To do so, on the package.json file it was needed just to modify the husky configuration as follows:
```
"husky": {
    "hooks": {
      "commit-msg": "commit-message-validator",
      "pre-commit": "ng lint && ng test --watch=false --browsers=ChromeHeadless"
    }
  }
```
**package-lock.json** is automatically generated for any operations where npm modifies either the node_modules tree, or **package.json**. It describes the exact tree that was generated, such that subsequent installs are able to generate identical trees, regardless of intermediate dependency updates.


package.json
```diff
@@ -100,7 +100,7 @@
   "husky": {
     "hooks": {
       "commit-msg": "commit-message-validator",
-      "pre-commit": "ng lint"
+      "pre-commit": "ng lint && ng test --watch=false"
```
scastillo-jp, line 103: Parece que no esta funcionando muy bien.

Dañe un par de tests y en la hook solo detecta el  **lint** pero no los test

`"pre-commit": "ng lint && ng test --watch=false"
`
 hice la prueba cambiando el  orden

`"pre-commit": "ng test --watch=false && ng lint"
`y solo detecta  los test y no el lint.

como recomendación sería bueno que uses en los test  --browsers=ChromeHeadless
algo así
`"pre-commit": "ng test --watch=false --browsers=ChromeHeadless & ng lint" `

DrFreud1: Thank you for your comments, Sandro. It worked when I was testing the changes on my computer, maybe the problem was that I did not commit the package-lock.json file which I know now that automatically changes when changing the package.json file.

I have tested it again, and it works as expected. It does not allow you to perform the commit if a test fails or if the lint fails.

package.json
```diff
@@ -80,7 +80,7 @@
     "commit-message-validator": "0.1.11",
     "datatables.net": "1.10.21",
     "datatables.net-dt": "1.10.21",
-    "husky": "4.2.3",
+    "husky": "^4.3.8",
```
ridouku, line 83: remove the caret symbol

ridouku (changes requested):

package.json
DrFreud1: Solved

ridouku (approved):

PR #727, opened 2021-08-18
fix: TT-312 [UI] Calendar the section titles is not responsive

ridouku:
### Observation

Actually the responsive in time entries is not working
Screen Shot 2021-08-17 at 17 13 10
There is a mistake with name Previus, and the titles aren’t responsive

Screen Shot 2021-08-17 at 17 13 10

### Solution

The scss of calendar component was changed to truncate the tittle text and fix time entries responsive, was changed Previus to Previous.

Screen Shot 2021-08-17 at 17 13 10
Screen Shot 2021-08-17 at 17 12 42
Screen Shot 2021-08-18 at 10 21 23


DrFreud1 (approved):
As far as I tested, the functionality is working fine and I do not see any problems with the implementation. Good job @ridouku !

PR #728, opened 2021-08-18
TT-304 Handle message: the data could not be load

edgardavid2015:
# Problem
In the time-entries section, when there is no data, the message "The data could not be load" is displayed.
Error-dataMessage
# Solution
When there is no data, instead of the message "The data could not be load", the message "No data available in table" should be displayed in the table section. That is the purpose of this pull request.
solution-errorMessage

### DIFF
--- a/src/app/modules/time-entries/pages/time-entries.component.html
+++ b/src/app/modules/time-entries/pages/time-entries.component.html
@@ -54,12 +54,12 @@
             <th class="col"></th>
           </tr>
         </thead>
-        <tr *ngIf="checkIfDataSourceIsLoadingorEmpty(dataSource)">
+        <tr *ngIf="dataSource.isLoading">
           <td class="text-center" colspan="7">{{NO_DATA_MESSAGE}}</td>
         </tr>
         <app-loading-bar *ngIf="dataSource.isLoading"></app-loading-bar>
         <tbody *ngIf="!dataSource.isLoading">
-          <tr *ngIf="!checkIfDataSourceIsLoadingorEmpty(dataSource)">
+          <tr *ngIf="!dataSource?.data.length">
             <td class="text-center" colspan="7">{{NO_DATA_MESSAGE}}</td>
           </tr>
           <tr class="d-flex" *ngFor="let entry of dataSource.data">
--- a/src/app/modules/time-entries/pages/time-entries.component.html
+++ b/src/app/modules/time-entries/pages/time-entries.component.html
@@ -59,7 +59,7 @@
         </tr>
         <app-loading-bar *ngIf="dataSource.isLoading"></app-loading-bar>
         <tbody *ngIf="!dataSource.isLoading">
-          <tr *ngIf="checkIfDataSourceIsLoadingorEmpty(dataSource)">
+          <tr *ngIf="!checkIfDataSourceIsLoadingorEmpty(dataSource)">
             <td class="text-center" colspan="7">{{NO_DATA_MESSAGE}}</td>
           </tr>
           <tr class="d-flex" *ngFor="let entry of dataSource.data">
--- a/src/app/modules/time-entries/pages/time-entries.component.html
+++ b/src/app/modules/time-entries/pages/time-entries.component.html
@@ -42,7 +42,7 @@
   </ng-template>
   <ng-template #listView>
     <div id="listView">
-      <table class="table table-sm table-striped mb-0" datatable [dtTrigger]="dtTrigger" [dtOptions]="dtOptions" *ngIf="timeEntriesDataSource$ | async as dataSource">
+      <table class="table table-sm table-striped mb-0" *ngIf="timeEntriesDataSource$ | async as dataSource">
         <thead class="thead-blue">
           <tr class="d-flex">
             <th class="col">Date</th>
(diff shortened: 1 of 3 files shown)


src/app/modules/time-entries/pages/time-entries.component.html
```diff
@@ -54,8 +54,14 @@
             <th class="col"></th>
           </tr>
         </thead>
+        <tr *ngIf="dataSource.isLoading">
+          <td class="text-center" colspan="7">No data available in table</td>
+        </tr>
         <app-loading-bar *ngIf="dataSource.isLoading"></app-loading-bar>
         <tbody *ngIf="!dataSource.isLoading">
+          <tr *ngIf="dataSource?.data.length === 0">
+            <td class="text-center" colspan="7">No data available in table</td>
```
ridouku, line 63: I think you can create a readonly variable to get this message, because you are repeating the same string

src/app/modules/time-entries/pages/time-entries.component.html
```diff
@@ -54,8 +54,14 @@
             <th class="col"></th>
           </tr>
         </thead>
+        <tr *ngIf="dataSource.isLoading">
```
ridouku, line 57: Could you make a function to check this, because you have two if condition that check if the element must show the message

ridouku (changes requested):

ridouku (approved):

src/app/modules/time-entries/pages/time-entries.component.html
edgardavid2015: Thank you for the review, I am going to apply this change.

src/app/modules/time-entries/pages/time-entries.component.html
edgardavid2015: You're right, that's a better approach. I am going to make that function, thank you.

src/app/modules/time-entries/pages/time-entries.component.ts
```diff
@@ -216,4 +218,18 @@ export class TimeEntriesComponent implements OnInit, OnDestroy {
       });
     }
   }
+
+  checkIfDataSourceIsLoading(data: DataSource<Entry>): boolean {
+    if (data.isLoading) {
+      return true;
+    }
+    return false;
```
jcalarcon98, line 226: Good Job Edgar,  you can improve this as follows:
```suggestion
   return data.isLoading;
```
However I also consider eliminating this method because it is irrelevant to return a property, but my recommendation differs from gustavo's, so you can keep the method, but improve it
**Note:** Take care if for some reason data is undefined or null, you can get the following error at execution time:
image

But if you are really sure that it will never be null or undefined, you can leave it as is

src/app/modules/time-entries/pages/time-entries.component.ts
```diff
@@ -216,4 +218,18 @@ export class TimeEntriesComponent implements OnInit, OnDestroy {
...
+  }
+
+  checkIfDataSourceIsEmpty(data: DataSource<Entry>): boolean {
+    if (data.data.length === 0) {
+      return true;
+    }
+    return false;
```
jcalarcon98, line 233: The same suggestion as above!

jcalarcon98 (commented):
Good job Edgar 💥 , only a few suggestions, if you consider implement it.

src/app/modules/time-entries/pages/time-entries.component.ts
edgardavid2015: Thank you for the revision, I am going to make changes to improve my code.

src/app/modules/time-entries/pages/time-entries.component.ts
edgardavid2015: Same explanation as above.

src/app/modules/time-entries/pages/time-entries.component.html
```diff
@@ -54,8 +54,14 @@
             <th class="col"></th>
           </tr>
         </thead>
+        <tr *ngIf="checkIfDataSourceIsLoadingorEmpty(dataSource)">
+          <td class="text-center" colspan="7">{{NO_DATA_MESSAGE}}</td>
+        </tr>
         <app-loading-bar *ngIf="dataSource.isLoading"></app-loading-bar>
         <tbody *ngIf="!dataSource.isLoading">
+          <tr *ngIf="!checkIfDataSourceIsLoadingorEmpty(dataSource)">
```
jcalarcon98, line 62: I'm not sure if this implementation is correct, for example in line `61`, you do this:
`<tbody *ngIf="!dataSource.isLoading">`
Here you check if dataSOurce is loading, and this extract of code only appear if dataSource loading has been completed (i.e. `false`), however in line `62`, that method `checkIfDataSourceIsLoadingorEmpty(dataSource)`, verifies again if dataSource is loading, I think that implementation is incorrect, because it does not make sense to perform the verification again if you know that dataSource.isLoading variable isi `false`

scastillo-jp (approved):

src/app/modules/time-entries/pages/time-entries.component.ts
```diff
@@ -216,4 +218,11 @@ export class TimeEntriesComponent implements OnInit, OnDestroy {
       });
     }
   }
+
+  checkIfDataSourceIsLoadingorEmpty(source?: DataSource<Entry>): boolean {
+    if (source.isLoading && source.data.length === 0) {
+      return true;
+    }
+    return false;
+  }
```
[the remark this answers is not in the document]
edgardavid2015: Thanks again for your review. I am going to delete this method.

src/app/modules/time-entries/pages/time-entries.component.html
edgardavid2015: Thanks for your review. Line 61 was not modified, and line 62 will have another approach.

DrFreud1 (approved):
I do not have any observations about the last changes. Besides, the functionality is working fine for me.

PR #729, opened 2021-08-18
feat: TT-314 Synchronize local projects with Server projects

jcalarcon98:
## Description

Currently there is the functionality to store recent projects in the LocalStorage, this feature works correctly, however, it does not keep an update with the data that exists on the server, that is, if a project/customer is deactivated and if the person had it saved in the LocalStorage of his browser, he will still have it available and will be able to mark entries with that project, even though the project/customer has already been deactivated.

## Solution
With this PR a synchronization is established between the user's projects in the LocalStorage and the projects returned by the server. For example, if the user has a project in the LocalStorage that has been disabled, it will be removed since it does not exist in the server.

### DIFF
--- a/src/app/modules/shared/utils/project-storage.util.spec.ts
+++ b/src/app/modules/shared/utils/project-storage.util.spec.ts
@@ -1,5 +1,6 @@
 import { updateProjectStorage, getProjectsOnStorage } from './project-storage.util';
 import { Project } from '../models/project.model';
+import { PROJECTS_KEY_FOR_LOCAL_STORAGE } from '../../../../environments/environment';
 
 describe('Project Storage', () => {
 
@@ -10,7 +11,7 @@ describe('Project Storage', () => {
   let localStorageGetItemMock;
 
   beforeEach(() => {
-    projectsIdentifier = 'projectsSelected';
+    projectsIdentifier = PROJECTS_KEY_FOR_LOCAL_STORAGE;
 
     testProject = {
       customer: {
--- a/src/app/modules/shared/utils/project-storage.util.spec.ts
+++ b/src/app/modules/shared/utils/project-storage.util.spec.ts
@@ -6,32 +6,28 @@ describe('Project Storage', () => {
   let storageProjects: Project[];
   let projectsIdentifier: string;
   let serverProjects: Project[];
+  let testProject: Project;
+  let localStorageGetItemMock;
 
   beforeEach(() => {
     projectsIdentifier = 'projectsSelected';
 
+    testProject = {
+      customer: {
+        name: 'ioet Inc. (was E&Y)',
+      },
+      id: 'f3630e59-9408-497e-945b-848112bd5a44',
+      name: 'Time Tracker',
+      customer_id: '20c96c4d-5e26-4426-a704-8bdd98c83319',
+      status: 'active',
+    };
+
     storageProjects = [
-      {
-        customer: {
-          name: 'ioet Inc. (was E&Y)',
-        },
-        id: 'f3630e59-9408-497e-945b-848112bd5a44',
-        name: 'Time Tracker',
-        customer_id: '20c96c4d-5e26-4426-a704-8bdd98c83319',
-        status: 'active',
-      }
+      testProject
     ];
 
     serverProjects = [
-      {
-        customer: {
-          name: 'ioet Inc. (was E&Y)',
-        },
-        id: 'f3630e59-9408-497e-945b-848112bd5a44',
-        name: 'Time Tracker',
-        customer_id: '20c96c4d-5e26-4426-a704-8bdd98c83319',
-        status: 'active',
-      },
+      testProject,
       {
         customer: {
           name: 'No Matter Name',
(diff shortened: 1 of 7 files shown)


src/app/modules/shared/utils/project-storage.util.ts
```diff
@@ -0,0 +1,30 @@
+import { Project } from '../models/project.model';
+import { PROJECTS_KEY_FOR_LOCAL_STORAGE } from '../../../../environments/environment';
+
+const projectsKey = PROJECTS_KEY_FOR_LOCAL_STORAGE;
```
ridouku, line 4: I think is not necessary define a constant to this, because you can use directly the value in your code, if you want change the name you could use import as to change the name or define it as readonly inside your function

src/app/modules/shared/utils/project-storage.util.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export function updateProjectStorage(serverProjects: Project[]): void {
+  const storageProjects: Project[] = getProjectsOnStorage(projectsKey);
+  const isServerProjectsNotEmpty = serverProjects && serverProjects.length !== 0;
```
ridouku, line 8: you could use lodash library and use isEmpty utility to validate both cases, you can review about this library here https://lodash.com and the implementation here https://www.npmjs.com/package/lodash

src/app/modules/shared/utils/project-storage.util.ts
```diff
@@ -0,0 +1,30 @@
...
+    });
+
+    const projectsForLocalStorage = JSON.stringify(updatedStorageProjects);
+    localStorage.setItem(projectsKey, projectsForLocalStorage);
+  }
+}
+
+export function getProjectsOnStorage(projectsIdentifier: string): Project[] {
+  const projectsInsideLocalStorage: string = localStorage.getItem(projectsIdentifier);
+  return projectsInsideLocalStorage && JSON.parse(projectsInsideLocalStorage);
```
ridouku, line 27: Why are you returning two different types?

src/environments/environment.ts
```diff
@@ -33,6 +33,14 @@ export const ROLES = {
     value: 'time-tracker-tester',
   },
 };
+
+/*
+This variable is used by the ng-autocomplete component used in these files:
+- details-fielfs.component.ts
```
ridouku, line 39: It's not necessary to put all files where variable is used, remove this

src/app/modules/shared/utils/project-storage.util.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    serverProjects = [
+      {
+        customer: {
+          name: 'ioet Inc. (was E&Y)',
+        },
+        id: 'f3630e59-9408-497e-945b-848112bd5a44',
+        name: 'Time Tracker',
+        customer_id: '20c96c4d-5e26-4426-a704-8bdd98c83319',
+        status: 'active',
```
ridouku, line 33: when you mock something and it is repeating you must define a variable to avoid the repetition of values, for example here you are using the same values in two objects,

src/app/modules/shared/utils/project-storage.util.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    updateProjectStorage(serverProjects);
+
+    expect(localStorage.setItem).toHaveBeenCalledWith(projectsIdentifier, JSON.stringify([]));
+  });
+
+  it('If Server projects is empty, should not update the localStorage', () => {
+    serverProjects = [];
+    spyOn(localStorage, 'getItem').and.returnValue(JSON.stringify(storageProjects));
+    spyOn(localStorage, 'setItem');
```
ridouku, line 70: I think you could put the spyOn in the method beforeEach or create a function that receive the values to set in the spyOn and then call it in each test, because you are are repeating the same mocks in three tests

ridouku (changes requested):

src/app/modules/shared/utils/project-storage.util.ts
ridouku, line 27: Here you can use lodash to get the value, I think is more clear when you use the next `getItem(object, 'path', defaultValue)` and you can prevent errors

ridouku (changes requested):

src/app/modules/shared/utils/project-storage.util.ts
jcalarcon98: That's a good suggestion Gustavo, I forgot the as inside the import, I will implement in that form, just a clarification, as you can see we are using functions, and the readonly is only available at the class level

src/app/modules/shared/utils/project-storage.util.ts
jcalarcon98: Good suggestion Gustavo, it was implemented only because lodash is already used in the project, however, for such a simple verification that can be performed with the base language, I consider unnecessary the implementation of the library, besides adding a dependency to it in case you want to replace it in the future.

src/environments/environment.ts
jcalarcon98: This was a suggestion made by Sandro

src/app/modules/shared/utils/project-storage.util.spec.ts
jcalarcon98: Perfect, it was overlooked, now that is implemented.

src/app/modules/shared/utils/project-storage.util.spec.ts
jcalarcon98: Perfect, it was overlooked, now that is implemented.

src/app/modules/shared/utils/project-storage.util.ts
jcalarcon98: Same comment as above. There are two possibilities described when using `localStorage.getItem()`, you can return the value as such or just null, see here. https://developer.mozilla.org/en-US/docs/Web/API/Storage/getItem, so I think it is unnecessary to use lodash for this case.

ridouku (approved):

scastillo-jp (approved):

JosueOb:
I reviewed the changes you made and tested them in the development environment with test data, using the script to generate test data in the local database. Test entries were generated and some projects were disabled. As a result, the projects were no longer presented in the Time Clock and Add new entry forms and removed from local storage. Very good job @jcalarcon98.

PR #735, opened 2021-08-23
feat: TT-321 Improve create/edit Entry modal

jcalarcon98:
## Description
The user experience in the create/edit input modal was improved, more intuitive icons and a button to drag the modal has been added, the result is as follows:
image

### DIFF
--- a/src/app/modules/time-entries/pages/time-entries.component.html
+++ b/src/app/modules/time-entries/pages/time-entries.component.html
@@ -99,11 +99,11 @@
       <div class="modal-header">
         <h5 class="modal-title">{{ entryId ? 'Edit Entry' : 'New Entry' }}</h5>
         <div>
-          <button class="btn shadow-none"  data-bs-toggle="tooltip" title="Clean form values"  (click)="detailsFields.cleanFieldsForm()">
-            <i class="fa fa-eraser"></i>
+          <button class="btn shadow-none"  data-bs-toggle="tooltip" title="Clean form values" (click)="detailsFields.cleanFieldsForm()">
+            <em class="fa fa-eraser"> </em>
           </button>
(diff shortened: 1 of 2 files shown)


src/app/modules/time-entries/pages/time-entries.component.html
```diff
@@ -96,11 +96,16 @@
...
         <h5 class="modal-title">{{ entryId ? 'Edit Entry' : 'New Entry' }}</h5>
-        <button type="button" class="btn" (click)="detailsFields.cleanFieldsForm()">
-          <i class="fa fa-undo" aria-hidden="true"></i>
-        </button>
+        <div>
+          <button class="btn shadow-none"  data-bs-toggle="tooltip" title="Clean form values" (click)="detailsFields.cleanFieldsForm()">
+            <em class="fa fa-eraser"> </em>
+          </button>
+          <button class="btn shadow-none"  data-bs-toggle="tooltip" title="Drag modal" cdkDragHandle>
+            <em class="fa fa-grip-vertical"></em>
```
ridouku, line 106: I think that icon is incorrect, because is like a drop down icon, I mean some users could confuse it with a menu options, you could replace it with a hand icon like this
image

ridouku (changes requested):

src/app/modules/time-entries/pages/time-entries.component.html
jcalarcon98: Thanks for your suggestion 👍🏼  @ridouku, this icon is used by Github to drag and drop the repositories, you can see the in your github profile, however, a common decision has already been reached with the team.

DrFreud1 (approved):
It is working well on its functionality, and in my opinion there are not issues with the implementation.

ridouku (approved):

edgardavid2015 (approved):
Despite the different opinions about the drag icon, the rest of the changes works fine, excellent job @jcalarcon98.

PR #740, opened 2021-08-26
Tt 318 improve column visibility section

edgardavid2015:
# Problem
In the Reports section, when we select the "Column visibility" button, this button shows a button row where each button represents a column name in the table.
This is working fine, but there is no notable difference between the selected buttons and no selected buttons inside this button row.

column_visibility_before

Moreover, when there is a reduction in the screen size, the "Column visibility" button is not showing anything.

column_visibility_screen_size_reduction

# Solution
In this pull request when we select the "Column visibility" button, we can see a clear difference between the selected buttons and no selected buttons by their color.

column_visibility_after

Also, when the screen size changes, the "Column visibility" button will display the button row as it normally does, adding a scroll in the y-axis.

column_visibility_screen_size_after

**Note:**
**As a bonus, the print button functionality was improved, now it only prints the columns that are displayed in the table.**

### DIFF
--- a/src/app/modules/reports/components/time-entries-table/time-entries-table.component.ts
+++ b/src/app/modules/reports/components/time-entries-table/time-entries-table.component.ts
@@ -33,13 +33,6 @@ export class TimeEntriesTableComponent implements OnInit, OnDestroy, AfterViewIn
       {
         extend: 'excel',
         exportOptions: {
-          format: {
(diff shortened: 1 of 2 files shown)


JosueOb:
I have tested the last two commits and the Excel and CSV options show the contents of the selected columns. Well done Edgar.

PR #744, opened 2021-09-01
fix: TT-328 when editing/creating entries set seconds to 0

DrFreud1:
With the modification of the lines of code of this pull request, each new time entry created from the "Time Entries" menu will have the following hour format established: "HH:mm:ss", where ss will always be set to 00 to avoid conflicts with inexistent time overlaps.

### DIFF
--- a/src/app/modules/time-entries/pages/time-entries.component.spec.ts
+++ b/src/app/modules/time-entries/pages/time-entries.component.spec.ts
@@ -387,7 +387,7 @@ describe('TimeEntriesComponent', () => {
       uri: 'http://testing.is.fun',
       activity_id: 'sss',
       project_id: 'id',
-      start_date: new Date(),
+      start_date: new Date(new Date().setSeconds(0)),
       end_date: new Date(new Date().setHours(0, 0, 0, 0))
     };
     state.timeEntriesDataSource.data = [lastEntry];
--- a/src/app/modules/time-entries/pages/time-entries.component.spec.ts
+++ b/src/app/modules/time-entries/pages/time-entries.component.spec.ts
@@ -385,7 +385,6 @@ describe('TimeEntriesComponent', () => {
     const currentDate = new Date();
     currentDate.setSeconds(defaultSeconds);
     currentDate.setMilliseconds(defaultSeconds);
-    const emptyDate = new Date(new Date().setHours(0, 0, 0, 0));
     const lastEntry = {
       description: 'testing is fun',
       technologies: [],
@@ -393,7 +392,7 @@ describe('TimeEntriesComponent', () => {
       activity_id: 'sss',
       project_id: 'id',
       start_date: currentDate,
-      end_date: emptyDate
+      end_date: currentDate
     };
     state.timeEntriesDataSource.data = [lastEntry];
     mockEntriesSelector = store.overrideSelector(getTimeEntriesDataSource, state.timeEntriesDataSource);
--- a/src/app/modules/time-entries/pages/time-entries.component.spec.ts
+++ b/src/app/modules/time-entries/pages/time-entries.component.spec.ts
@@ -381,14 +381,18 @@ describe('TimeEntriesComponent', () => {
   it('should preload data of last entry when a project is selected while creating new entry ', waitForAsync(() => {
     component.entry = null;
     component.entryId = null;
+    const defaultSeconds = 0;
+    const currentDate = new Date();
+    currentDate.setSeconds(defaultSeconds);
+    const emptyDate = new Date(new Date().setHours(0, 0, 0, 0));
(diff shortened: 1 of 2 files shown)


src/app/modules/time-entries/pages/time-entries.component.spec.ts
```diff
@@ -387,7 +387,7 @@ describe('TimeEntriesComponent', () => {
       uri: 'http://testing.is.fun',
       activity_id: 'sss',
       project_id: 'id',
-      start_date: new Date(),
+      start_date: new Date(new Date().setSeconds(0)),
```
jcalarcon98, line 390: I think that this is not so clear when reading the code. Maybe, you can improve like:
```suggestion
      const defaultSeconds = 0
      const currentDate = new Date();
      currentDate.setSeconds(0);
      
      start_date: currentDate,
```

src/app/modules/time-entries/pages/time-entries.component.ts
```diff
@@ -146,7 +146,7 @@ export class TimeEntriesComponent implements OnInit, OnDestroy {
           uri: dataToUse.uri ? dataToUse.uri : '',
           activity_id: dataToUse.activity_id,
           project_id: dataToUse.project_id,
-          start_date: new Date(),
+          start_date: new Date(new Date().setSeconds(0)),
```
jcalarcon98, line 149: The same suggestion as above.

jcalarcon98 (commented):
Good Job @DrFreud1, you can check the suggestions I made, if you consider, implement it

src/app/modules/time-entries/pages/time-entries.component.ts
DrFreud1: Thank very much for your feedback @jcalarcon98, I really appreciate that. I have implemented your suggestions, but with some other minor changes I considered necessary.

src/app/modules/time-entries/pages/time-entries.component.spec.ts
```diff
@@ -381,14 +381,18 @@ describe('TimeEntriesComponent', () => {
...
+    const emptyDate = new Date(new Date().setHours(0, 0, 0, 0));
     const lastEntry = {
       description: 'testing is fun',
       technologies: [],
       uri: 'http://testing.is.fun',
       activity_id: 'sss',
       project_id: 'id',
-      start_date: new Date(),
-      end_date: new Date(new Date().setHours(0, 0, 0, 0))
+      start_date: currentDate,
```
scastillo-jp, line 394: Verify if currentDate has millisecond in 0.

src/app/modules/time-entries/pages/time-entries.component.ts
```diff
@@ -139,15 +139,18 @@ export class TimeEntriesComponent implements OnInit, OnDestroy {
...
+        const emptyDate = new Date(new Date().setHours(0, 0, 0, 0));
         const entry = {
           description: dataToUse.description ? dataToUse.description : '',
           technologies: dataToUse.technologies ? dataToUse.technologies : [],
           uri: dataToUse.uri ? dataToUse.uri : '',
           activity_id: dataToUse.activity_id,
           project_id: dataToUse.project_id,
-          start_date: new Date(),
-          end_date: startDate
+          start_date: currentDate,
```
scastillo-jp, line 152: Check here first if the current date has milliseconds at 0, then update the test if necessary.

DrFreud1:
> Check here first if the current date has milliseconds at 0, then update the test if necessary. 

Thank you for your review Sandro, with the next commit, this is fixed

src/app/modules/time-entries/pages/time-entries.component.ts
```diff
@@ -139,15 +139,19 @@ export class TimeEntriesComponent implements OnInit, OnDestroy {
     this.store.pipe(select(getTimeEntriesDataSource)).subscribe(ds => {
       const dataToUse = ds.data.find(item => item.project_id === event.projectId);
       if (dataToUse && this.isNewEntry()) {
-        const startDate = new Date(new Date().setHours(0, 0, 0, 0));
+        const defaultSeconds = 0;
+        const currentDate = new Date();
+        currentDate.setSeconds(defaultSeconds);
+        currentDate.setMilliseconds(defaultSeconds);
+        const emptyDate = new Date(new Date().setHours(0, 0, 0, 0));
```
jcalarcon98, line 146: GIven that currentDate is the same that emptyDate, we can remove emptyDate and use currentDate for end_date too.

jcalarcon98 (approved):
I Review this functionality and all works perfectly, Good Job @DrFreud1 💥, Only my last suggestion at level code, if you consider implement it.

PR #750, opened 2021-09-10
TT-331 redesign of the sidebar

edgardavid2015:
# Problem
The current sidebar of the time tracker application is OK, it's working and has no functionality errors. But this sidebar could have a better design that enhances the user experience.

current_sidebar

# Solution
In this pull request, the sidebar has a modern design, with new icons, new colors, and new functionality.

new_sidebar


scastillo-jp (approved):

JosueOb (changes requested):
Hi Edgar, I have checked your changes and everything works fine but I see that you are not using the color palette set for the project and you are not considering the changes for the dark mode.

src/app/modules/shared/components/sidebar/sidebar.component.scss
```diff
@@ -65,7 +79,7 @@ body {
 }
 
 .dark-mode-toggle {
```
JosueOb, line 81: This style is not used, it would be better to remove it.

JosueOb (commented):
I have finished reviewing the changes made and I have some suggestions, the first one is about the icons, I see that some SVG icons are still kept in the assets directory, I think those icons that are not used should be removed, and the position of the icon to show or hide the sidebar does not keep its initial position.

JosueOb (approved):

src/app/modules/shared/components/sidebar/sidebar.component.scss
edgardavid2015: Thanks for your review Josue, I will delete it in the next commit.

PR #753, opened 2021-09-17
refactor: TT-327 Replace recently projects

EdansRocks:
## Problem
Currently the ng-autocomplete component is used to select the projects in the Time Clock and Time entries section, which stores in Local Storage the history of recent projects that have been selected, which generates problems because it may happen that a project is deactivated but as it is stored in Local Storage the users can still do Clock In. Previously a patch was made to synchronize these projects in Local Storage with the server projects.
image

## Solution
An endpoint was created in the backend to obtain the recent projects of each user according to the last 20 entries, given this the solution is to replace the ng-autocomplete component that uses the Local Storage to consume the information from the backend endpoint, replacing the previous functionality.


JosueOb (approved):
I've already reviewed the changes made and tested them in the local environment with test data in the database to test their functionality.  So I was able to get the projects from an endpoint and be submitted in the input of the time entry form. Good work @EdansRocks .

eguezgustavo (approved):

PR #756, opened 2021-09-24
TT-354 fix bug in Add new activity button

edgardavid2015:
# Problem
When the "Add a new activity" button is clicked, it should reset the "Name" and "Description" fields.

activity-button-before

# Solution
This pull request implements the reset form fields function on the "Add new activity" button.

activity-button-now


DrFreud1 (approved):
I don't find any problems with the implementation.

JosueOb (approved):

PR #766, opened 2021-11-05
feat: TT-395 remove button create, update and dropdown

ararcos:
**Problem**
The activity view needs to be read-only to retrieve this information from json.
image

**Solution**
Hidden the modify and create buttons inside the activity view.
image


scastillo-jp:
Algunas recomendaciones:
- cambiar el Operations a Status
- el icono circular bajarlo un poco para que se quede alineado horizontal con el texto Active o inactive.

Screen Shot 2021-11-05 at 16 36 51

scastillo-jp (approved):

josepato87 (approved):

josepato87 (approved):

PR #770, opened 2021-12-02
fix: TT-413 show projects when you do not have the latest projects

ararcos:
**Problem**
Whenever a user enters in the Time Clock>Projects, they have to type in a project for the feature to show a message.

image

the message of the technologies is not clear and new users do not understand how it works
image

**Solution**
when it is a new user, all available projects are displayed and when it is an old user, only the old projects are displayed.

image

improved help text of technologi
image

### DIFF
--- a/src/app/modules/shared/components/technologies/technologies.component.ts
+++ b/src/app/modules/shared/components/technologies/technologies.component.ts
@@ -17,7 +17,7 @@ export class TechnologiesComponent implements OnInit, OnDestroy {
   readonly ALLOW_SELECT_MULTIPLE = true;
   readonly ALLOW_SEARCH = true;
   readonly MIN_SEARCH_TERM_LENGTH = 2;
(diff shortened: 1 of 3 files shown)


src/app/modules/shared/components/technologies/technologies.component.ts
```diff
@@ -17,7 +17,7 @@ export class TechnologiesComponent implements OnInit, OnDestroy {
   readonly ALLOW_SELECT_MULTIPLE = true;
   readonly ALLOW_SEARCH = true;
   readonly MIN_SEARCH_TERM_LENGTH = 2;
-  readonly TYPE_TO_SEARCH_TEXT = 'Please enter 2 or more characters';
+  readonly TYPE_TO_SEARCH_TEXT = 'Please enter 2 or more characters to search for technologies ';
```
mandres2015, line 20: It is better is rename to TYPE_TO_SEARCH_TECHNOLOGIES, because it cannot be used for other search term

mandres2015 (approved):

PR #772, opened 2021-12-03
fix: TT-437 validated maximum time entry hours

ararcos:
**Problem**
Whenever a user records a time entry, user has to record the date in and date out.
image

**Solution**
when the user enters a new time entry he can only enter the entry date and the exit date will be updated automatically.
image
image


mandres2015 (approved):

gancoce94 (approved):
NA, but check if report works well, this related to problem reported by Roland

PR #784, opened 2021-12-29
fix: TT-485 switch time entry

gancoce94:
Solve problem when switch from an active time entry, to another in a new project.


santii90ec (approved):
Reviewed

epulla (approved):

ararcos (commented):
check why you are failing the tests

PR #796, opened 2022-01-14
fix: TT-508 update endpoints in the UI

scastillo-jp:
The task TT-484 have changes to resolve problems with the Serverless routes and is necessary update in the UI end points with the param ID.

List of services to update in the UI
deleteActivity.
updateActivity
updateProject
deleteProject
deleteProjectType
updateProjectType
deleteCustomer
updateCustomer
updateEntry
deleteEntry


scastillo-jp:
This PR is not necessary, we have a better solution for the Backend, I'm go to close this PR

PR #797, opened 2022-01-20
Tt 501 integrate reports with the new backend

OscarPallazhco:
In this PR the endpoint created for reports is used. In production it will not change the route.

Working in local
Screenshot from 2022-01-20 13-59-28

Working in production
Screenshot from 2022-01-20 13-29-51


madinya (approved):
Approved!

epulla (approved):

PR #801, opened 2022-01-28
fix: TT-362-pagination-problem-solved

rofarinango:
Modification to fix the pagination error when change month with no data to month with data. Pagination is now keep displaying.


LuisMS7:
Review the codecov/path process and why it is failing.

LuisMS7 (approved):

bryanAguilar001 (approved):

PR #819, opened 2022-03-08
feat: TT-509 Use terraform to provision ui infraestructure

jatobrun:
Create the infraestructure folder with the 4 main files
(main.tf, prod.tfvars, stage.tfvars, variables.tf) and use
the organization private modules to deploy all the resources to
the ui need to deploy in the two enviroments (stage and prod)

Resolves: https://ioetec.atlassian.net/browse/TT-509


scoello (commented):
Looks very good. Please add the readme.md file

madinya (approved):
LGTM. keep in mind that you should also request a review from Fausto.

scoello (approved):
Good job my friend!

madinya (commented):
Reviewed with few comments

infrastructure/main.tf
```diff
@@ -0,0 +1,67 @@
...
+  workspace = terraform.workspace
+  config = {
+    resource_group_name  = "ioet-infra-tf-state"
+    storage_account_name = "timetrackertfstate"
+    container_name       = "time-tracker-tf-state"
+    key                  = "this.tfstate"
+  }
+}
+
+data "azurerm_container_registry" "registry" {
```
faustocv, line 35: **issue**: I think this will make another unnecessary call to the azure API. The `data.terraform_remote_state.service.outputs.container_registry_name` already provides a valid reference to the container registry

infrastructure/main.tf
```diff
@@ -0,0 +1,67 @@
...
+  name                = data.terraform_remote_state.service.outputs.container_registry_name
+  resource_group_name = data.terraform_remote_state.service.outputs.resource_group_name
+}
+
+data "azurerm_resource_group" "root" {
+  name = data.terraform_remote_state.service.outputs.resource_group_name
```
faustocv, line 41: **issue**: Same issue as the one explained above

infrastructure/main.tf
```diff
@@ -0,0 +1,67 @@
...
+}
+
+locals {
+  common_name             = "time-tracker-ui"
+  environment             = terraform.workspace
+  service_name            = "${local.common_name}-${local.environment}"
+  create_app_service_plan = true
```
faustocv, line 48: **suggestion**: I think this does not matter a lot as the image_name. Like this https://github.com/ioet/time-tracker-v2/blob/255112b0f270ae17de84a20cd7791705dc573f8e/infrastructure/main.tf#L54
Could you replace create_app_service_plan with image_name?

infrastructure/main.tf
```diff
@@ -0,0 +1,67 @@
...
+  service_plan_kind       = "Linux"
+}
+
+module "ui" {
```
faustocv, line 52: **issue**: I don't see the image_tag definition here. That one is relevant to associate it with git releases. Take a look at https://github.com/ioet/time-tracker-v2/blob/255112b0f270ae17de84a20cd7791705dc573f8e/infrastructure/main.tf#L73 as a reference

infrastructure/variables.tf
```diff
@@ -0,0 +1,17 @@
+variable "docker_image_name" {
```
faustocv, line 1: **issue:** replace this with image_tag, something like https://github.com/ioet/time-tracker-v2/blob/255112b0f270ae17de84a20cd7791705dc573f8e/infrastructure/variables.tf#L7

infrastructure/main.tf
jatobrun: same comment in the module dont have a tag definition so the image name change because the tag changes

infrastructure/main.tf
jatobrun: really? i change that

faustocv (approved):
Looks good now

PR #824, opened 2022-03-14
TT-582 Improve the summary section of the time-clock page and the clock out button

LuisMS7:


iHackN3WTON (approved):
LGTM, just may be round the corners on the current class orange box, to follow the style of the whole form.

PR #825, opened 2022-03-14
feat: TT-577 get profile pic from Google account

kjduy:


Olucas7 (approved):

src/app/modules/login/services/login.service.ts
```diff
@@ -20,7 +20,9 @@ export class LoginService {
   }
 
   signIn() {
-    this.socialAuthService.signIn(GoogleLoginProvider.PROVIDER_ID);
+    this.socialAuthService.signIn(GoogleLoginProvider.PROVIDER_ID).then(user => {
+      localStorage.setItem('googleUserPhoto', user.photoUrl);
```
rofarinango, line 24: Check warning on this line from Codecov, apparently this line was not covered by tests? Idk if its necessary to include a test on this since coverage right now I believe its on 95.64% on master branch.

PR #826, opened 2022-03-15
feat: TT-576 Improve grids visual presentation and function

iHackN3WTON:


codigodehoy (approved):
It looks like great, but I recommend the arrow's color should be another color because, the title is already white and doesn't distinguishable

madinya (approved):
LGTM

PR #839, opened 2022-03-29
TT-593 On Reports page, add filter by user name

codigodehoy:
* Created user-search component
* Loaded all users to the select
* sent the user.id to the api reports


MarcoAguirre (approved):

listerineh (approved):
Nice and understandable solution.

PR #928, opened 2022-08-29
TTA 115 time tracker doesnt notifies the user when they lost internet connection or have a slow connection

abigailscl:


mmaquina (changes requested):

src/app/modules/shared/services/status-network.service.ts
```diff
@@ -0,0 +1,33 @@
...
+@Injectable({
+  providedIn: 'root'
+})
+export class StatusNetworkService {
+
+  constructor(
+    private toastrService: ToastrService
+  ) { }
+
+  checkTypeError(dataError: ErrorType){
```
mmaquina, line 19: le ponemos de nombre swhowMessage?

mmaquina (approved):

PR #930, opened 2022-09-01
TTA-102-prevent-an-admin-to-revoke-its-own-credentials

mmaquina:


nicolsss (approved):

mmaquina:
Switch disabled to prevent admin from auto-revoke

PR #935, opened 2022-09-30
Tta 174 create pop up message for redirecting users from legacy to production

andresacg30:
Route redirection for legacy when prod is up :)


reihtw (commented):
There are some SonarCloud errors

reihtw (approved):
Good work!

PR #973, opened 2023-01-20
Added new credentials for azure login

mmaquina:


nishedcob (approved):
We made this change after reading the documentation found here: https://github.com/marketplace/actions/azure-login

